PR #1, opened 2015-09-04
Make GDB compile in native mode

palmer-dabbelt:
It looks like the merge of the native GDB patches didn't go correctly and ended up producing something that doesn't even build.  This PR submits some patches that make the RISC-V native GDB port build.  I haven't tested anything (and the last patch is a mess), but I thought I'd get this out there just in case anyone was working on it because I might not have time to finish this for a bit.

At least it can't be worse than code that doesn't build... :)


vapier:
looks good, thanks!

PR #4, opened 2016-03-09
make riscv64-unknown-elf build again

timsifive:
I needed these changes in order for the build to pass, after running `./configure --target=riscv64-unknown-elf`.
With these changes I can open gdb, and `show architecture` confirms it's using riscv.


vapier:
these seem to all align with changes i made in latest mainline gdb.  thanks!

PR #5, opened 2016-03-15
allow access to all CSRs

timsifive:
Every CSR is accessible as a named register, csrNN. 'info registers all' and 'info registers system' only show CSRs that have aliases defined. They are still accessible using their aliases (eg. mcause, mscratch).
    
This change renumbers the CSR register numbers used when talking to a remote client, breaking CSR accessibility for existing RISC-V gdbserver/stub implementations.


timsifive:
There's a bug here that breaks 'info all-registers' (which is subtly different from 'info registers all,' which does work).
I'll fix it and send a new pull request.

vapier:
this giant hand maintained list of registers is preciously the sort of thing we want to avoid

timsifive:
Can you clarify what exactly you want to avoid? The list of csr0 -- csr4095? Would you be happy if it was generated at run-time? Or do you object to the list existing in memory at all?

timsifive:
FWIW, I had taken a stab at an implementation that didn't rely on a list of all CSRs at all, but ran into trouble when it looked like we're reading registers through a cache that assumes regnums are contiguous, so would need ~5k entries anyway. It's possible there's a better solution but I'm not familiar with gdb internals.

vapier:
any version that requires hundreds of lines of static strings is bad.  if it could be dynamically generated, that might be interesting, but i'd still rather just move directly to XML.  having large XML db's isn't a problem.

timsifive:
So you would be OK listing all 4096 csr registers in the XML description?

vapier:
yeah, dumping large db's into XML is fine

timsifive:
OK. I'll see how much work it is to use the XML approach.

timsifive:
I couldn't find any documentation on the XML approach. Looking at some ARM and MIPS cores, it turns this:
```
  <reg name="r0" bitsize="32" type="uint32"/>
  <reg name="r1" bitsize="32" type="uint32"/>
```

Into:
```
  tdesc_create_reg (feature, "r0", 0, 1, NULL, 32, "uint32");
  tdesc_create_reg (feature, "r1", 1, 1, NULL, 32, "uint32");
```

Is that really an improvement? It looks like the giant list still exists as C code, but now there's an extra step in building it.

I had assumed that the XML file gets loaded at run-time, so it's possible to change the definitions without recompiling gdb. That doesn't appear to be the case. (Or I'm not looking in the right place.)

timsifive:
Ah, there's "set tdesc" which does allow XML files to be read at runtime.

PR #6, opened 2016-03-19
Allow access to all CSRs through $csrNN, and access to GPRs using $xN

timsifive:
The reason for this change is to allow a gdb user to access arbitrary CSR registers that may be implemented on a given core. In 'info all-registers', 'info registers all', and 'info registers system' only the CSRs that have aliases defined are displayed. To keep the code simple, this change changes the register numbers gdb uses internally and when communicating to clients using its own protocol. Any existing clients will need to be modified to accommodate the new numbers.

It's also possible now to access GPRs using both their "simple name" (eg. x10) as well as their alias name (eg. a0).

This implementation doesn't use XML files, because I couldn't find the mechanism documented anywhere. I believe the new code is a strict improvement over the existing code, and doesn't make it any harder to switch to XML files at a later time.


timsifive:
PS
This implementation generates CSR register names each time they're required, instead of relying on a large list.

timsifive:
What would it take to get this pull request accepted?
It fixes a few small bugs, adds functionality that will be required in the future, and doesn't break anything AFAIK.

palmer-dabbelt:
I'm working on toolchain stuff this week: I'm going to go through the riscv-gnu-toolchain bugs and fix the easy ones, rebase this onto master again, merge in all the patches, and then point riscv-gnu-toolchain here.  Then I'll look at this PR, I don't want this repo to fork.

timsifive:
OK. Please let me know when you're done with your side of things. Then I'll go through and fix conflicts in this pull request.

gdb/riscv-tdep.c
```diff
@@ -65,7 +65,7 @@ struct riscv_frame_cache
   struct trad_frame_saved_reg *saved_regs;
 };
 
-static const char * const riscv_gdb_reg_names[RISCV_NUM_REGS] =
+static const char * const riscv_gdb_reg_names[RISCV_LAST_FP_REGNUM+1] =
```
vapier: style: should be a space around the +

gdb/riscv-tdep.c
```diff
@@ -156,6 +152,9 @@ static const struct register_alias riscv_register_aliases[] =
   { "ft9", 62 },
   { "ft10", 63 },
   { "ft11", 64 },
+#define DECLARE_CSR(name, num) { #name, (num)+65 },
```
vapier: style: should be a space around the +

gdb/riscv-tdep.c
```diff
@@ -630,14 +614,24 @@ riscv_register_reggroup_p (struct gdbarch  *gdbarch,
 {
   int float_p;
   int raw_p;
+  unsigned int i;
+
+  /* Used by 'info registers' and 'info registers <groupname>'. */
```
vapier: style: there should be two spaces after the period.  this comes up a few times.

gdb/riscv-tdep.c
```diff
@@ -646,14 +640,55 @@ riscv_register_reggroup_p (struct gdbarch  *gdbarch,
...
+          return 1;
+    }
+    return 0;
+  } else if (reggroup == vector_reggroup)
     return 0;
   else
     internal_error (__FILE__, __LINE__, _("unhandled reggroup"));
 }
 
+static void
```
vapier: did you really need to relocate this function ?  would be better to keep it where it was.

gdb/riscv-tdep.c
```diff
@@ -1157,8 +1193,14 @@ riscv_gdbarch_init (struct gdbarch_info  info,
       tdesc_data = tdesc_data_alloc ();
 
       valid_p = 1;
-      for (i = RISCV_ZERO_REGNUM; i < RISCV_LAST_REGNUM; ++i)
-	valid_p &= tdesc_numbered_register (feature, tdesc_data, i, riscv_gdb_reg_names[i]);
+      for (i = RISCV_ZERO_REGNUM; i <= RISCV_LAST_FP_REGNUM; ++i)
+        valid_p &= tdesc_numbered_register (feature, tdesc_data, i,
+            riscv_gdb_reg_names[i]);
```
vapier: pretty sure indentation is off -- it should be lined up using tabs (and then spaces) to the "feature" above

gdb/riscv-tdep.c
```diff
@@ -1157,8 +1193,14 @@ riscv_gdbarch_init (struct gdbarch_info  info,
...
+      for (i = RISCV_FIRST_CSR_REGNUM; i <= RISCV_LAST_CSR_REGNUM; ++i)
+        {
+          sprintf(buf, "csr%d", i - RISCV_FIRST_CSR_REGNUM);
```
vapier: style: need a space before the (

if buf is only used here, move the decl from above to this scope

gdb/riscv-tdep.h
```diff
@@ -71,14 +71,14 @@ enum {
...
+#define RISCV_FIRST_CSR_REGNUM (RISCV_LAST_FP_REGNUM + 1)
+#define DECLARE_CSR(name, num) RISCV_ ## num ## _REGNUM = RISCV_LAST_FP_REGNUM + 1 + num,
 #include "opcode/riscv-opc.h"
 #undef DECLARE_CSR
-#define RISCV_LAST_CSR_REGNUM RISCV_LAST_REGNUM - 1
+#define RISCV_LAST_CSR_REGNUM (RISCV_LAST_REGNUM - 1)
 
   /* Leave this as the last enum.  */
-  RISCV_LAST_REGNUM
+  RISCV_LAST_REGNUM = RISCV_LAST_FP_REGNUM+4097
```
vapier: please add/use an existing define rather than inlining magic constants like "4097"

vapier:
it's an interesting approach.  can you clean up your branch though ?  squash/throw away the old/unwanted changes.

gdb/riscv-tdep.c
timsifive: Relocating allowed me to not add a prototype for it. Would you like me to move it back and add the prototype?

gdb/riscv-tdep.c
timsifive: I assume you mean that "riscv_gdb_reg_names" on line 1198 should start at the same column as "feature" on line 1197.
Done.

gdb/riscv-tdep.c
timsifive: Done.

gdb/riscv-tdep.c
timsifive: Done.

gdb/riscv-tdep.c
timsifive: Done.

gdb/riscv-tdep.c
timsifive: I assume you want a space between "sprintf" and ( on line 1201. Done.
I moved buf.

gdb/riscv-tdep.h
timsifive: Changed the code a bit so now it's all enum definitions again.

timsifive:
Can you outline how I would squash the unwanted changes? I can't simply rewrite the history since I already published to github. At least that's how I understand it.
(Really it sounds like a lot of work for an insignificant benefit, but I'm willing to jump through the hoops if you can guide me through it.)

timsifive:
What's the status of this pull request? Are you waiting on me to do something more?

palmer-dabbelt:
Like vapier said, it would be good to clean up the branch because it's a bit of a mess.

timsifive:
I couldn't figure out how to change this pull request to use a different branch, so I created https://github.com/riscv/riscv-binutils-gdb/pull/9 which is identical to this request, except the history has been cleaned up.

PR #9, opened 2016-05-21
Allow access to all CSRs through $csrNN, and access to GPRs using $xN (cleaner history)

timsifive:
The reason for this change is to allow a gdb user to access arbitrary CSR registers that may be implemented on a given core. In 'info all-registers', 'info registers all', and 'info registers system' only the CSRs that have aliases defined are displayed. To keep the code simple, this change changes the register numbers gdb uses internally and when communicating to clients using its own protocol. Any existing clients will need to be modified to accommodate the new numbers.

It's also possible now to access GPRs using both their "simple name" (eg. x10) as well as their alias name (eg. a0).

This implementation doesn't use XML files, because I couldn't find the mechanism documented anywhere. I believe the new code is a strict improvement over the existing code, and doesn't make it any harder to switch to XML files at a later time.

This request is identical to https://github.com/riscv/riscv-binutils-gdb/pull/6, but its history has been cleaned up a bit.

### DIFF
--- a/gdb/riscv-tdep.c
+++ b/gdb/riscv-tdep.c
@@ -188,15 +188,18 @@ value_of_riscv_user_reg (struct frame_info *frame, const void *baton)
 }
 
 static const char *
-riscv_register_name (struct gdbarch *gdbarch,
-		     int             regnum)
+register_name (struct gdbarch *gdbarch,
+	       int             regnum,
+               int             prefer_alias)
 {
   int i;
   static char buf[20];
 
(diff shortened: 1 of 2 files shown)


aswaterman:
@palmer-dabbelt @vapier looks like this new PR addresses your concerns from #6 -- is it good to go?

palmer-dabbelt:
This is merged as https://github.com/riscv/riscv-binutils-gdb/commit/0efb5fd05efd127d33ffd44cff8feba5a516990f .  Thanks!

PR #12, opened 2016-07-26
32-bit addresses should not be sign-extended.

timsifive:
That's a MIPS thing.
This fixes gdb debugging 32-bit code where bit 31 of the address is set.


aswaterman:
Ah, I see.

PR #14, opened 2016-08-09
Riscv binutils 2.26 freebsd

bukinr:


ld/emultempl/riscvelf.em
```diff
@@ -39,7 +43,16 @@ riscv_elf_before_allocation (void)
...
+  /* check if all of input files are binary blob.
+     If true, I skip relaxation pass. */
+  for (abfd = link_info.input_bfds; abfd != NULL; abfd = abfd->link.next) {
+    if (abfd->xvec != &binary_vec) {
+      all_inputs_binary = FALSE;
+      break;
+    }
+  }
+
+  link_info.relax_pass = all_inputs_binary == TRUE? 0: 2;
```
aswaterman: This seems like a hack.  Shouldn't this be done in _bfd_riscv_relax_section, skipping each individual pass if binary?  Comparing against &binary_vec also seems brittle.

In relax_section, there's a check that bails out if reloc_count == 0.  Why isn't that triggered for binary sections?

gas/configure.tgt
```diff
@@ -393,7 +393,8 @@ case ${generic_target} in
   ppc-*-kaos*)				fmt=elf ;;
   ppc-*-lynxos*)			fmt=elf em=lynx ;;
 
-  riscv*-*-*)			fmt=elf endian=little em=linux ;;
+  riscv*-linux-*)			fmt=elf endian=little em=linux ;;
+  riscv*-*-freebsd*)			fmt=elf endian=little em=freebsd ;;
```
aswaterman: why different patterns for linux and freebsd?

ld/emultempl/riscvelf.em
y-shibat: Hi, Andrew,
I wrote a original fix for freebsd binutils branch.
I also think this is a kind of dirty hack.  
The riscv-gnu-toolchain-freebsd branch does not contain all of binutils sources. So I chose to fix anything as much as possible in sources left on the branch. If the bsd suppose is fully merged to the mainline, I think we can implement it in more cleaner way.

> In relax_section, there's a check that bails out if reloc_count == 0. Why isn't that triggered for binary sections?
Although I don't remember details exactrly, I suppose the value of link_info.relocatable is checked in bfd_generic_relax_sections() in bfd/reloc.c. and its called before the vail out by reloc_count ==0.
# Now I come back to my family home without bringing my development environment and memo.

Anyway, I'll try to write new patch that is for the riscv binutils mainline after I go back to my house next week.

PR #15, opened 2016-08-10
Fix wrong behavior for divw and remw

kito-cheng:
- According spec result is -1 is only when divisor is -1 AND dividend
   is -2^31 for divw.

 - According spec result is 0 is only when divisor is -1 AND dividend
   is -2^31 for divw.

### DIFF
--- a/sim/riscv/sim-main.c
+++ b/sim/riscv/sim-main.c
@@ -653,9 +653,11 @@ execute_m (SIM_CPU *cpu, unsigned_word iw, const struct riscv_opcode *op)
   const char *rs1_name = riscv_gpr_names_abi[rs1];
   const char *rs2_name = riscv_gpr_names_abi[rs2];
   unsigned_word tmp, dividend_max;
+  signed_word dividend32_max;
   sim_cia pc = cpu->pc + 4;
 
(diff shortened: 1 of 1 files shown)


palmer-dabbelt:
@kito-cheng Thanks for the fix!  I believe your commit message is wrong, but the code is good.  Do you mind fixing the message so I don't get confused?  If I'm reading everything correctly, the result of "divw -2^31 / -1" is "-2^31" (and "remw -2^31 / -1" is "0").

It looks like this matches a test case in riscv-tests <https://github.com/riscv/riscv-tests/blob/master/isa/rv64um/divw.S>.  Have you considered running riscv-tests on sim?  There's test cases for most of the RISC-V edge cases in there (and this way I don't have to go read the spec to make sure you're correct).

palmer-dabbelt:
OK, I'm closing this just because the words are wrong (and I want to bump to 2.27).  See

https://github.com/riscv/riscv-binutils-gdb/pull/16

vapier:
you can edit/update PR's fwiw

i ran all the available isa tests at the time when i wrote the sim, but it was a pita to get the tests to build in a usable format ...

palmer-dabbelt:
@vapier I only know how to update PRs from branches I can push to, but for forks I just close and re-open them.  I tried Google and didn't find anything more than I know, am I missing something?

PR #16, opened 2016-08-10
Fix wrong behavior for divw and remw

palmer-dabbelt:
- According spec result is -2^31 is only when divisor is -1 AND dividend
   is -2^31 for divw.

 - According spec result is 0 is only when divisor is -1 AND dividend
   is -2^31 for remw.


vapier:
thanks!

PR #17, opened 2016-08-15
Fix the low-half of PC-relative relocations

palmer-dabbelt:
@a0u found the bug.  I just messed up when going through and changing everything so the various callbacks changed to be void.


aswaterman:
haha, this is almost like goto-fail

PR #20, opened 2016-09-02
fixing compilation error with older linux env

ekrimer:
In my environment CLOCK_BOOTTIME is missing. For this case I suggest to revert to CLOCK_MONOTONIC. 
A little poking on this topic revealed that this could be a reason for platform incompatibility. According to different forums OSX doesn't have clock_gettime (I don't use OSX so couldn't confirm), other OSes would have different enums for clock IDs. My CentOS is happy with CLOCK_MONOTONIC but perhaps worth a more robust solution, especially if someone can try on other OSes. Until then would be good to have this hack in since riscv-tools build fails without it.

### DIFF
--- a/sim/riscv/sim-main.c
+++ b/sim/riscv/sim-main.c
@@ -90,7 +90,12 @@ fetch_csr (SIM_CPU *cpu, const char *name, int csr, unsigned_word *reg)
       {
 	struct timespec ts;
 
+#if defined(CLOCK_BOOTTIME)
 	if (clock_gettime (CLOCK_BOOTTIME, &ts) == 0)
+#else
+	if (clock_gettime (CLOCK_MONOTONIC, &ts) == 0)
+#endif
+
 	  {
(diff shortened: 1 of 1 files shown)


aswaterman:
Related: is using any clock a good idea?  It will make the simulator nondeterministic.  A deterministic alternative would be to make it linear in instructions retired.

aswaterman:
Regardless, this is an improvement, so I'd suggest merging -- @palmer-dabbelt ok?

palmer-dabbelt:
This is fine, but I'd also be OK using a fake clock for the simulator like we do in spike.  Note that our port is the only sim port that relies on librt (which is necessary for this clock stuff on old glibc versions).

While you shouldn't need to assign copyright for this change (it's too small), if you plan on doing future toolchain development you should get the copyright assignment paperwork for the FSF started sooner rather than later.  If you're not sure how to do that, just email me.

Thanks for the code!

Issue #22, opened 2016-09-13
break dependency on Linux


sbeamer:
The riscv simulator depends on Linux's clock_gettime, which prevents this from being compiled on any other platform such as OS X, even if gcc is used. Since this repo is a dependency of riscv-gnu-toolchain, this restriction makes any other OS incapable of using riscv-gcc.

clock_gettime is used to implement CSR_TIME (in sim/riscv/sim-main.c), but it seems any timing mechanism could also be used. Some possibilities include:
- gettimeofday() - Yes, its time can be changed, but does this matter for simulation?
- C++11 chrono library
- [Hack](http://stackoverflow.com/a/17112198/2548418) just for OS X

Are there are other constraints? Could one of the above suggestions work?

aswaterman:
I would favor not depending on the current time at all; it makes the simulator nondeterministic.  Could just use an affine function of the instructions-retired counter.

sbeamer:
Could it be the same as instructions-retired? Maybe just instructions-retired x 2?

aswaterman:
something like that would probably be ok

aswaterman:
(in spike, they are equal)

vapier:
`clock_gettime` is not Linux specific.  it's explicitly in POSIX:
  http://pubs.opengroup.org/onlinepubs/9699919799/functions/clock_getres.html

we can fallback to `CLOCK_MONOTONIC` if `CLOCK_BOOTTIME` is not available.  `getimeofday` is not appropriate as it's not monotonic -- it can be rewound.  any C++ lib is not appropriate as the GNU sim is C only, and doesn't rely on external C++ libs.

if none of those are available because the OS is crap, we can fallback to `csr.instret`.

aswaterman:
Trying all three is a reasonable approach, too... @vapier what do you think about the tradeoff between having an accurate time and having deterministic simulator runs?

palmer-dabbelt:
This is fixed, CSR_TIME now returns the cycle count.

PR #23, opened 2016-09-13
make CSR_TIME return cycles

sbeamer:
In reference to #22, by tracking cycles can add determinism and remove dependence on clock_gettime. In the current sim implementation, cycles is incrementing equivalently to instructions retired.

### DIFF
--- a/sim/riscv/Makefile.in
+++ b/sim/riscv/Makefile.in
@@ -31,6 +31,5 @@ SIM_OBJS = \
 	interp.o \
 	machs.o \
 	sim-main.o
-SIM_EXTRA_LIBS = -lrt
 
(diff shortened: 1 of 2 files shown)


aswaterman:
I'm cool with this approach

sbeamer:
Bump @aswaterman. Is this blocking on anything?

palmer-dabbelt:
We're doing some repo restructuring, but I like this commit so I'm going to try not to lose it.

palmer-dabbelt:
https://github.com/riscv/riscv-binutils-gdb/commit/d96aaeb86be2dc78089ed8a59db0a1990268b501

PR #25, opened 2016-10-08
Leave decr_pc_after_break at 0.

timsifive:
It was set to 4, which tells gdb that when a breakpoint is hit the
reported PC might be 4 after the breakpoint. AFAICT this is not the case
in RISC-V, and it is certainly not the case when debugging through
OpenOCD. In fact this change causes a problem when debugging through
OpenOCD if there is a software breakpoint set at foo, and a hardware
trigger set at foo+4. When the hardware trigger is hit, gdb will
decrement the PC by 4, find the software breakpoint, and then change the
PC on the target to match the software breakpoint. It's never possible
to resume execution, and the instruction that has the software
breakpoint on it is executed every time the user tries.

I don't know how this affects native RISC-V gdb, but Palmer says: "Well,
I'd just speculate it's a GDB bug and submit the PR, ideally with a test
case of some sort.  As far as I know nobody a really runs native GDB,
but if they do then hopefully they'll speak up and we'll figure out
what's going on."

### DIFF
--- a/gdb/riscv-tdep.c
+++ b/gdb/riscv-tdep.c
@@ -1203,7 +1203,6 @@ riscv_gdbarch_init (struct gdbarch_info  info,
   /* Information about the target architecture.  */
   set_gdbarch_return_value (gdbarch, riscv_return_value);
   set_gdbarch_breakpoint_from_pc (gdbarch, riscv_breakpoint_from_pc);
-  set_gdbarch_decr_pc_after_break (gdbarch, 4);
   set_gdbarch_remote_breakpoint_from_pc (gdbarch, riscv_remote_breakpoint_from_pc);
   set_gdbarch_print_insn (gdbarch, print_insn_riscv);
 


palmer-dabbelt:
Tim and I talked about this a bit.  Neither of us is 100% sure this is the right thing to do, and I don't have a working RISC-V Linux environment set up to test anything with.

@a0u @vapier @sagark do any of you guys know if this is correct?  If not, do you have a working setup to test on?  Even just riscv-linux and riscv-tools hashes that both build would be useful.

sagark:
I don't have a setup for native GDB, but GDB remote debugging can be used with QEMU (both system mode and user mode) following this:
https://github.com/riscv/riscv-qemu#using-qemu-to-debug-risc-v-code

QEMU works with https://github.com/riscv/riscv-tools/releases/tag/priv-1.9-rc0 and the priv-1.9 branch of riscv-linux

timsifive:
On Mon, Oct 10, 2016 at 10:53 AM, Sagar Karandikar <notifications@github.com
> wrote:

> I don't have a setup for native GDB, but GDB remote debugging can be used
> with QEMU (both system mode and user mode) following this:
> https://github.com/riscv/riscv-qemu#using-qemu-to-debug-risc-v-code
>

That's not working for me. I'm using the bblvmlinuxinitramfs_dynamic image
which boots into bash just fine, but when I connect with gdb (with or
without -S), gdb requests a list of registers ('g') and never gets a
response. (There is a little bit of regular connection setup that happens
before this which does complete correctly.) I'm using
qemu 8e35f13f02b68089476e9a018d03d0a955155dff.

Tim

sagark:
You need to use a newer version of qemu, gdb support was re-enabled by the commit after that one (2ef31d248d16fa489ba86bd92727de5b5b9766ab).

timsifive:
Thanks, that mostly works. I should’ve done a ‘git pull’ before complaining.

So I’m just attaching to the kernel I have and doing some single steps, and
setting breakpoints to run to. The main weird thing I’m observing (with
unchanged gdb) is that when I single step over a csrr/csrw instruction,
qemu steps that instruction and the next one as well. Ie. when debugging:

   0x0000000080000220 <+120>:    li    t6,0
   0x0000000080000224 <+124>:    csrw    mscratch,zero
   0x0000000080000228 <+128>:    auipc    t0,0x0
   0x000000008000022c <+132>:    addi    t0,t0,-544 # 0x80000008 <trap_vector>
   0x0000000080000230 <+136>:    csrw    mtvec,t0
   0x0000000080000234 <+140>:    csrr    t1,mtvec
   0x0000000080000238 <+144>:    bne    t0,t1,0x80000238 <do_reset+144>
   0x000000008000023c <+148>:    auipc    sp,0x8ec
   0x0000000080000240 <+152>:    addi    sp,sp,1803 # 0x808ec947

Single step from 0x80000224 results in the PC being on 0x8000022c, and t0
modified.
Single step from 0x80000230 results in the PC being on 0x8000023c (3
instructions were executed!).

Just from manually smoke testing the behavior is identical regardless of
whether decr_pc_after_break is 0 or 4.

Tim
​

On Mon, Oct 10, 2016 at 12:58 PM, Sagar Karandikar <notifications@github.com

timsifive:
So given that it doesn't affect qemu debugging (AFAICT), can this change be merged?

palmer-dabbelt:
It's merged, but github didn't detect it:

  https://github.com/riscv/riscv-binutils-gdb/commit/d29facb185f5442c2b7343c203034d14b2fd1297

I'm closing the PR.

Thanks for the patch!

timsifive:
Thanks, Palmer!

Tim

PR #27, opened 2016-10-28
Update mstatus to match 1.9.1 spec.

timsifive:


aswaterman:
Code looks good, but needs to adhere to GNU guidelines (every 8 spaces = 1 tab; braces on next line and indented 2 spaces).  Or ditch the braces.

aswaterman:
thanks!

PR #41, opened 2016-12-14
Use cbreak software breakpoint if supported.

timsifive:
This makes debugging code that uses the compressed ISA work.


gdb/riscv-tdep.c
```diff
@@ -158,35 +158,44 @@ static const struct register_alias riscv_register_aliases[] =
...
+    return 2;
+  } else {
+    return 4;
+  }
 }
 
 static const gdb_byte *
 riscv_sw_breakpoint_from_kind (struct gdbarch *gdbarch, int kind, int *size)
 {
   /* TODO: Support C.EBREAK for compressed (16-bit) insns.  */
```
palmer-dabbelt: @timsifive I think you deleted the wrong comment?

gdb/riscv-tdep.h
```diff
@@ -69,10 +69,13 @@ enum {
 
 #define RISCV_LAST_REGNUM (RISCV_NUM_REGS - 1)
 
+typedef enum { UNKNOWN, YES, NO } supported_t;
```
palmer-dabbelt: I'm a bit worried about namespace conflicts here, as these are all common words and this is a header file.  Is GDB C++ code now?  If so, can you make this an "enum class"?

gdb/riscv-tdep.c
timsifive: I omitted to delete the cbreak comment.
I intended to delete the long instruction one, because no special handling is needed for those. Just overwite the start of the instruction with whatever size breakpoint and everything will just work.

gdb/riscv-tdep.h
timsifive: I don't feel great about relying on C++ so soon after the transition. How about:
```
typedef enum { SUP_UNKNOWN, SUP_YES, SUP_NO } supported_t;
```

palmer-dabbelt: Works for me.

gdb/riscv-tdep.c
```diff
@@ -158,35 +158,44 @@ static const struct register_alias riscv_register_aliases[] =
...
+    /* TODO: Because we try to read misa, it is not possible to set a
+     * breakpoint before connecting to a live target. A suggested workaround is
+     * to look at the ELF file in this case. */
+    struct frame_info *frame = get_current_frame ();
+    uint32_t misa = get_frame_register_unsigned (frame, RISCV_CSR_MISA_REGNUM);
+    if (misa & (1<<2))
+      gdbarch_tdep (gdbarch)->supports_compressed_isa = YES;
+    else
+      gdbarch_tdep (gdbarch)->supports_compressed_isa = NO;
+  }
```
aswaterman: Need to use the GNU coding style:

    if (foo)
      {
        statements;
      }

gdb/riscv-tdep.c
```diff
@@ -158,35 +158,44 @@ static const struct register_alias riscv_register_aliases[] =
...
+
+  if (gdbarch_tdep (gdbarch)->supports_compressed_isa == YES) {
+    return 2;
+  } else {
+    return 4;
+  }
```
aswaterman: Just delete the braces on this one, or use a ternary operator.

gdb/riscv-tdep.c
```diff
@@ -158,35 +158,44 @@ static const struct register_alias riscv_register_aliases[] =
...
+
+  *size = kind;
+  switch (kind) {
+    case 2:
+      return c_ebreak;
+    case 4:
+      return ebreak;
+    default:
+      gdb_assert(0);
+  }
```
aswaterman: This one should be even weirder:

    switch
      {
      case 2:
        return c_ebreak;

aswaterman:
Not a fan of the brace style, but they'll reject the patches otherwise :|

timsifive:
I think I addressed your comments, but origin/riscv-next went all wonky on me. It doesn't even have gdb/riscv-tdep.c anymore. What's going on?

aswaterman:
@timsifive we rebased onto the head of binutils master, where gdb is not upstreamed.  @palmer-dabbelt can you add gdb back to riscv-next?

palmer-dabbelt:
This is merged.  I did it manually because github really went off the rails.

b1da828669b6a75f43e8dd3f48430a217fd88d84

PR #42, opened 2016-12-16
Support aligning text section from odd addresses

aswaterman:
Previously, the alignment directives were not correctly supported
in the text section when current alignment was only 1 byte (i.e.,
when the address was odd).  Since there are no 1-byte instructions
in RISC-V, this patch resolves the bug by writing a zero byte to
obtain 2-byte alignment, at which point a 2-byte NOP can be used
to obtain 4-byte alignment.

Resolves https://github.com/riscv/riscv-gnu-toolchain/issues/205


aswaterman:
@palmer-dabbelt this needs to go upstream

aswaterman:
Rebased on riscv-next.

palmer-dabbelt:
https://github.com/riscv/riscv-binutils-gdb/commit/3c20e8c0a1f4d5bc9f695fd2e6d10cbbd7b328f5

PR #43, opened 2016-12-16
Avoid creating symbol table entries for registers

aswaterman:
Instructions like "jal t0, foo" were erroneously creating symbol table
entries for t0 as well as foo, which causes linking problems.  Fix by
reordering instruction alternatives so that t0 is first attempted to
be parsed as a register, rather than as a symbol.


aswaterman:
This needs to go upstream, as it's necessary to build ld.so correctly.

aswaterman:
Rebased on riscv-next.

palmer-dabbelt:
https://github.com/riscv/riscv-binutils-gdb/commit/200d8d8f639fbb69b0a8df3da55d8b34fc100107

PR #44, opened 2016-12-19
Fix build failure.

timsifive:
It still doesn't build, but it gets further in the process. Now I get
```
make[3]: Leaving directory '/home/tnewsome/SiFive/riscv-tools/riscv-gnu-toolchain/riscv-binutils-gdb/gdb'
make[2]: *** No rule to make target '../sim/riscv/libsim.a', needed by 'gdb'.  Stop.
make[2]: Leaving directory '/home/tnewsome/SiFive/riscv-tools/riscv-gnu-toolchain/riscv-binutils-gdb/gdb'
Makefile:10454: recipe for target 'all-gdb' failed
make[1]: *** [all-gdb] Error 2
make[1]: Leaving directory '/home/tnewsome/SiFive/riscv-tools/riscv-gnu-toolchain/riscv-binutils-gdb'
Makefile:849: recipe for target 'all' failed
make: *** [all] Error 2
```

I'm not sure how to fix that one.


palmer-dabbelt:
Two things:

* This builds for me, and was suggested by one of the code reviewers.  I'm on a2, which is one of the Berkeley Ubuntu machines.  They've all got a bunch of non-standard stuff on them, what does your setup look like?
* The sim error is probably some missing run of "autoreconf -i", possibly in the sim directory.

timsifive:
I'm running Debian GNU/Linux 8 (jessie), gcc 4.9.2.

The problem I was fixing was that gcc complained that const was being discarded.

loc4_frag is declared const on line 2064 of gas/config/tc-riscv.c. Then on line 2076 it's passed as the first argument to fix_new_exp(). That function's prototype is
```
extern fixS *fix_new_exp
  (fragS * frag, int where, int size, expressionS *exp, int pcrel,
   bfd_reloc_code_real_type r_type);
```
in line 183 of gas/write.h.

Is your code different, is your gcc just not emitting a warning for this problem, or is there something else going on?

aswaterman:
I agree, it's hard to see how this would have built on any version of gcc.

palmer-dabbelt:
OK, it's actually just a warning on a2.  Sorry for breaking this.

I'll merge this PR (and the other ones) once I get my upstream branch sanitized WRT riscv-next, as I want to make sure I didn't lose anything.

palmer-dabbelt:
Your patch is here <https://github.com/riscv/riscv-binutils-gdb/commit/e80ebbb4f920dd3a6234b9a458dd8bb3c28e7e74>, I also re-ran the autoconf stuff.  Hopefully it all builds now!

PR #45, opened 2016-12-21
Implement Q extension

kito-cheng:


aswaterman:
@palmer-dabbelt do we want to pull this upstream before 2.28?

palmer-dabbelt:
Looks good to me.  I'll send it up.

palmer-dabbelt:
This is on riscv-next, so I'm closing it.

PR #48, opened 2016-12-30
Add missing instruction for Q extension: fcvt.d.q and fcvt.q.d

kito-cheng:


palmer-dabbelt:
Thanks!

aswaterman:
are you sending this upstream?

palmer-dabbelt:
I'm going to squash it into the other one and send a set upstream once we get the relro thing worked out.

PR #51, opened 2017-01-03
Make the PC and other GPRs unsigned.

timsifive:
Fixes https://github.com/sifive/openocd/issues/8

Pros of unsigned: addresses with the high bit set work as expected, eg. in `x/i $pc`.
Pros of signed: values that actually are signed are printed as expected, eg. in `p $t0`.


palmer-dabbelt:
I merged this manually.

PR #56, opened 2017-02-08
Don't use "_gp" on RISC-V, use "_global_pointer$" instead

palmer-dabbelt:
"_gp" could conflict with ABI-complient code.  While it's probably OK
because MIPS uses this name, we figured it'd be good to clean this up
before a release with RISC-V in it.

ld/ChangeLog:

2017-02-08  Palmer Dabbelt  <palmer@dabbelt.com>

        * emulparams/elf32lriscv-defs.sh (SDATA_START_SYMBOLS): Change
        _gp to __global_pointer$.

bfd/ChangeLog:

2017-02-08  Palmer Dabbelt  <palmer@dabbelt.com>

        * elfnn-riscv.c (riscv_global_pointer_value): Change _gp to
        __global_pointer$.

### DIFF
--- a/bfd/elfnn-riscv.c
+++ b/bfd/elfnn-riscv.c
@@ -1463,7 +1463,7 @@ riscv_global_pointer_value (struct bfd_link_info *info)
 {
   struct bfd_link_hash_entry *h;
 
-  h = bfd_link_hash_lookup (info->hash, "_gp", FALSE, FALSE, TRUE);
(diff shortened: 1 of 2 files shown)


palmer-dabbelt:
@aswaterman We forgot about this.  I can talk to upstream and see if it's OK, and I'll go run all the tests.  I think we might as well try it.

aswaterman:
Oops.  Good point.

Don't forget to change sysdeps/riscv/start.S in glibc, too.

PR #57, opened 2017-02-08
Compressed breakpoint option

timsifive:
This allows people to force the gdb to use or not use compressed
breakpoints, in case misa isn't accessible to determine this
automatically. Specifically, there's a user whose core does not support
compressed instructions but misa is not where gdb expects it to be
because it was moved in the 1.9.1 privileged spec.


palmer-dabbelt:
Thanks!

PR #58, opened 2017-02-13
Check the reserved high-order bits in e_flags on RISC-V

palmer-dabbelt:
It looks like we made one of the more embarrassing interface design
mistakes computer architects can make: not checking reserved bits.  On
RISC-V, e_flags contains ABI information that prevents users from
linking incompatible objects together.  This patch ensures the
unallocated flag bits are zero when linking, which allows us to safely
use these flags later.

bfd/ChangeLog

2017-02-13  Palmer Dabbelt  <palmer@dabbelt.com>

        * elfnn-riscv.c (_bfd_riscv_elf_merge_private_bfd_data): Check
        the reserved bits in e_flags for ibfd and obfd.

include/ChangeLog

2017-02-13  Palmer Dabbelt  <palmer@dabbelt.com>

        * riscv.h (EF_RISCV_RESERVED): Define.

### DIFF
--- a/bfd/ChangeLog
+++ b/bfd/ChangeLog
@@ -1,5 +1,10 @@
 2017-02-13  Palmer Dabbelt  <palmer@dabbelt.com>
 
+	* elfnn-riscv.c (_bfd_riscv_elf_merge_private_bfd_data): Check the
+	reserved bits in e_flags for ibfd and obfd.
+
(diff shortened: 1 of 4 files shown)


palmer-dabbelt:
@aswaterman I haven't tested this, but it does fix the last non-GDB-related issue on our github.

aswaterman:
roffle.  good catch, merge once it bootstraps.

PR #62, opened 2017-02-23
Add new counter-enable CSRs

aswaterman:


palmer-dabbelt:
https://sourceware.org/ml/binutils/2017-02/msg00300.html

palmer-dabbelt:
This landed upstream.

PR #63, opened 2017-03-02
Fix debug info

kito-cheng:
https://github.com/riscv/riscv-gnu-toolchain/issues/219

### DIFF
--- a/gas/config/tc-riscv.c
+++ b/gas/config/tc-riscv.c
@@ -1966,6 +1966,8 @@ md_apply_fix (fixS *fixP, valueT *valP, segT seg ATTRIBUTE_UNUSED)
 		  if (subtype < 0x80 && (subtype & 0x40))
 		    {
 		      /* DW_CFA_advance_loc */
+		      fixP->fx_frag = (fragS *) fixP->fx_frag->fr_opcode;
+		      fixP->fx_next->fx_frag = fixP->fx_frag;
 		      fixP->fx_r_type = BFD_RELOC_RISCV_SET6;
 		      fixP->fx_next->fx_r_type = BFD_RELOC_RISCV_SUB6;
 		    }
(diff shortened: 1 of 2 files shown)


gas/config/tc-riscv.c
```diff
@@ -1932,30 +1933,31 @@ md_apply_fix (fixS *fixP, valueT *valP, segT seg ATTRIBUTE_UNUSED)
 
 	    case BFD_RELOC_RISCV_CFA:
 	      /* Load the byte to get the subtype.  */
-	      subtype = bfd_get_8 (NULL, &fixP->fx_frag->fr_literal[fixP->fx_where]);
+	      subtype = bfd_get_8 (NULL, &((fragS *) (fixP->fx_frag->fr_opcode))->fr_literal[fixP->fx_where]);
+	      loc = fixP->fx_frag->fr_fix - (subtype & 7);
```
aswaterman: offsetT instead of long?

gas/config/tc-riscv.c
```diff
@@ -1847,6 +1847,7 @@ md_apply_fix (fixS *fixP, valueT *valP, segT seg ATTRIBUTE_UNUSED)
   unsigned int subtype;
   bfd_byte *buf = (bfd_byte *) (fixP->fx_frag->fr_literal + fixP->fx_where);
   bfd_boolean relaxable = FALSE;
+  long loc;
```
aswaterman: offsetT instead of long?

aswaterman:
@palmer-dabbelt want to try to sneak this upstream?

kito-cheng:
updated, use offsetT instead of long.

palmer-dabbelt:
Thanks for fixing this.  It looks like I missed the 2.28 release window, so I'm just going to submit it to master.

Issue #59, opened 2017-02-16
Enhance assembler/disassembler output for lui/c.lui


kito-cheng:
Arch: RV*C

**Test case 1 for disassembler:**
```
$ cat t1.s
x:
  li s5, -4096
$ riscv64-unknown-elf-gcc t1.s -c
$ riscv64-unknown-elf-objdump -d t1.o
0000000000000000 <x>:
   0:	7afd                	lui	s5,0xfffff
```
Expected result:
```
0000000000000000 <x>:
   0:	7afd                	lui	s5,-4096
```

**Test case 2 for disassembler:**

```
$ cat t4.s 
x:
  li s5, -4096
$ riscv64-unknown-elf-gcc t4.s -c -march=rv64ima
$ riscv64-unknown-elf-objdump -d t4.o 
0000000000000000 <x>:
   0:	fffffab7          	lui	s5,0xfffff
```
Expected result:
```
0000000000000000 <x>:
   0:	fffffab7          	lui	s5,-4096
```

**Test case 3 for assembler:**
```
$ cat t2.s 
x:
  c.lui s5, -4096
$ riscv64-unknown-elf-gcc t2.s -c
t2.s: Assembler messages:
t2.s:2: Error: illegal operands `c.lui s5,-4096'
```
Expected result:
Accept the input.

**Test case 4 for assembler:**
```
$ cat t3.s
x:
  lui s5, -4096
$ riscv64-unknown-elf-gcc t3.s -c -march=rv64ima
t3.s: Assembler messages:
t3.s:2: Error: lui expression not in range 0..1048575
```
Expected result:
Accept the input.

aswaterman:
```lui``` takes the immediate right-shifted by 12 and unsigned (i.e. 0x0 - 0xfffff), so -4096 would be 0xfffff.  I fear that changing this will cause problems.  In your proposal, what range of inputs would lui accept?

Seems better to rely on ```li``` if you want to use constants like -4096; this is what GCC does.

**Comment by kito-cheng** [comment]
Humm, I expect it `lui` can accept two kind of input, shifted immediate and signed integer.
For shifted immediate, accept 0xfffff000 instead of 0xfffff and output 0xfffff000 in disassembler.
For signed integer, accept all encode-able number.

And for disassembler, naybe we can add both kind of output?
``
0000000000000000 <x>:
   0:	fffffab7          	lui	s5,0xfffff000 # -4096
```

aswaterman:
[...]

PR #65, opened 2017-03-14
Fix assembler for c.li, c.andi, c.addi, c.addiw and c.lui

kito-cheng:
- User ISA spec 2.1 p86, dest in c.addi can be x0
- User ISA spec 2.1 p85, imm in c.li can be 0
- User ISA spec 2.1 p86, dest in c.addiw can be 0
- User ISA spec 2.1 p87, dest in c.andi can be 0
- User ISA spec 2.1 p85, imm in c.lui can accept 0x0~0x3f


gas/config/tc-riscv.c
```diff
@@ -1377,9 +1377,9 @@ riscv_ip (char *str, struct riscv_cl_insn *ip, expressionS *imm_expr,
 		  if (imm_expr->X_op != O_constant
 		      || imm_expr->X_add_number <= 0
 		      || imm_expr->X_add_number >= RISCV_BIGIMM_REACH
-		      || (imm_expr->X_add_number >= RISCV_RVC_IMM_REACH / 2
```
sorear: Is this actually correct?  "C.LUI loads the non-zero 6-bit immediate field into bits 17–12 of the destination register, clears the bottom 12 bits, and sign-extends bit 17 into all higher bits of the destination"  Sounds like a -32..31 to me, and qemu appears to implement it that way as well.

kito-cheng: Here is some discussion about lui: https://github.com/riscv/riscv-binutils-gdb/issues/59

aswaterman: I believe this is basically the same issue as #59.  While we made a bad decision about how to expose LUI, we are now stuck with it.

kito-cheng:
Patch updated :)

palmer-dabbelt:
4368ca4 Fix assembler for c.li, c.andi and c.addiw
4e9d526 Fix assembler for c.addi, rd can be x0

PR #68, opened 2017-03-31
RISC-V: Relax RISCV_PCREL_* to RISCV_GPREL_*

palmer-dabbelt:
In the medany code model the compiler generates PCREL_HI20+PCREL_LO12
relocation pairs against local symbols because HI20+LO12 relocations
can't reach high addresses.  We relax HI20+LO12 pairs to GPREL
relocations when possible, which is an important optimization for
Dhrystone.  Without this commit we are unable to relax
PCREL_HI20+PCREL_LO12 pairs to GPREL when possible, causing a XX%
permormance hit on Dhrystone on Rocket.

Note that we'll now relax

  la gp, __global_pointer$

to

  mv gp, gp

which probably isn't what you want in your entry code.  Users who want
gp-relative symbols to continue to resolve should add ".option norelax"
accordingly.  Due to this, the assembler now pairs PCREL relocations
with RELAX relocations when they're expected to be relaxed.

### DIFF
--- a/bfd/elfnn-riscv.c
+++ b/bfd/elfnn-riscv.c
@@ -2704,10 +2704,142 @@ riscv_relax_delete_bytes (bfd *abfd, asection *sec, bfd_vma addr, size_t count)
   return TRUE;
 }
 
+/* A second format for recording PC-relative hi relocations.  This stores the
+   information required to relax them to GP-relative addresses.  */
+
+typedef struct riscv_pcgp_hi_reloc riscv_pcgp_hi_reloc;
+struct riscv_pcgp_hi_reloc
+{
+  bfd_vma hi_sec_off;
+  bfd_vma hi_addend;
+  bfd_vma hi_addr;
+  unsigned hi_sym;
+  asection *sym_sec;
+  riscv_pcgp_hi_reloc *next;
+};
+
+typedef struct riscv_pcgp_lo_reloc riscv_pcgp_lo_reloc;
+struct riscv_pcgp_lo_reloc
+{
+  bfd_vma hi_sec_off;
+  riscv_pcgp_lo_reloc *next;
+};
+
+typedef struct
+{
+  riscv_pcgp_hi_reloc *hi;
+  riscv_pcgp_lo_reloc *lo;
+} riscv_pcgp_relocs;
+
+static bfd_boolean
+riscv_init_pcgp_relocs (riscv_pcgp_relocs *p)
+{
+  p->hi = NULL;
+  p->lo = NULL;
+  return TRUE;
+}
+
+static bfd_boolean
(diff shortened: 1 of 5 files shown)


bfd/elfnn-riscv.c
```diff
@@ -2704,10 +2704,167 @@ riscv_relax_delete_bytes (bfd *abfd, asection *sec, bfd_vma addr, size_t count)
...
+  riscv_pcgp_lo_reloc *next;
+};
+
+typedef struct
+{
+  riscv_pcgp_hi_reloc *hi;
+  riscv_pcgp_lo_reloc *lo;
+} riscv_pcgp_relocs;
+
+static bfd_boolean
```
aswaterman: why not void?

palmer-dabbelt:
This is broken, don't merge

palmer-dabbelt:
There's one remaining GCC test suite failure

````
gcc.dg/tree-prof/section-attr-1.c execution
````

**Comment by palmer-dabbelt** [comment]
@aswaterman This now boots Linux and passes the GCC test suite.  I've added R_RISCV_DELETE, which marks bytes for later deletion.  We should be able to use this to make the relaxations linear time, but I haven't actually done that yet.

**Comment by aswaterman** [comment]
R_RISCV_DELETE will never be emitted into an object file, right?  If so, we should do something like ```#define R_RISCV_DELETE (R_RISCV_max + 1)``` in ```elfnn-riscv.c```, rather than exposing it to the ELF ABI.  I believe this also means we don't need ```BFD_RELOC_RISCV_DELETE```.

**Comment by palmer-dabbelt** [comment]
Yes, that's better

**Comment by aswaterman** [comment]
I'm OK with merging this locally, but my intuition is that we should prototype the linear-time relaxation pass before upstreaming this.  I suspect that will reveal additional cobwebs.

**Comment by palmer-dabbelt** [comment]
Ya, that was next on my TODO list.  I'm going to clean this up and merge it for now, as I'm sick of needing to build binaries for people all the time.

File: include/elf/riscv.h [hunk 2]
```diff
@@ -89,6 +89,9 @@ START_RELOC_NUMBERS (elf_riscv_reloc_type)
   RELOC_NUMBER (R_RISCV_SET32, 56)
 END_RELOC_NUMBERS (R_RISCV_max)
 
+/* Relocation types not emitted to binaries. */
+#define R_RISCV_DELETE 127
+
```
aswaterman: Move to elfxx-riscv.h?

Did you pick 127 because of its ASCII meaning? :)

palmer-dabbelt:
I'm going to merge this, assuming it still passes the test suite.

aswaterman:
cool

PR #70, opened 2017-04-27
Change CALL macro to use ra as the temporary address register

michaeljclark:
e.g.

    1:  auipc ra, %pcrel_hi(symbol)
        jalr  ra, %pcrel_lo(1b)(ra)

The use of ra instead of t1 for address construction provides an opportunity for a microarchitecture to elide the write of the destination address, and instead read the target address as an immediate spread across the fused auipc+jalr pair. The link register ra in the jalr overwrites the target address temporary.

I've compiled the toolchain, built linux-kernel and boot tested it. Here is a snippet of objdump output after the change:

```
ffffffff80000134:	001b4097          	auipc	ra,0x1b4
ffffffff80000138:	6f6080e7          	jalr	1782(ra) # ffffffff801b482a <strcmp>
```


palmer-dabbelt:
https://sourceware.org/ml/binutils/2017-05/msg00008.html

PR #74, opened 2017-05-03
Fix disassemble for c.li, c.andi and c.addiw

kito-cheng:


aswaterman:
Oops :) can you write a gnu-style changelog entry?

kito-cheng:
Updated :)

palmer-dabbelt:
https://sourceware.org/ml/binutils/2017-05/msg00038.html

Issue #76, opened 2017-05-08
.eh_frame_hdr refers to overlapping FDEs when linking libc


palmer-dabbelt:
This is a regression from riscv-binutils-2.28 (3f21b5c9675db61ef5462442b6a068d4a3da8aaf) to riscv-next (9fc62d500cf36bb6c9d2360dfcd9397ddd417e4d).

````
[...]

PR #81, opened 2017-05-24
Prototype -mstrict-operands GAS option (RFC)

aswaterman:
As discussed in this thread: https://groups.google.com/a/groups.riscv.org/forum/?utm_medium=email&utm_source=footer#!msg/sw-dev/IHSRwpczMt4/xqOJrVAxBgAJ

@palmer-dabbelt @kito-cheng opinions?


aswaterman:
I haven't tested this thoroughly because of #76

aswaterman:
The other, unresolved debate is if we want to define ```add t0, t0, 1``` to be a valid alias of ```addi t0, t0, 1``` in the official assembly syntax.  Then, this PR is moot.  That remains my preference, but a vocal few on the mailing list disagree, and would prefer something along the lines of this PR.

sorear:
position: if this were proposed a year ago I would have enthusiastically pushed for -mstrict-operands to be the only option.  Today I'm very weakly opposed, because an optional simplification is not much of a simplification.

aswaterman:
@palmer-dabbelt adding a .option would be more symmetric with the others, but I don't think that granularity is useful.  Generally, the whole file will use conformant assembly, or none of it will.  Can you think of a case where allowing this on finer granularity would actually help?

aswaterman:
Also, adding a .option adds more stuff to the RISC-V assembly languagespec , where adding a flag only adds stuff to GAS.

michaeljclark:
Attached is the binutils patch backported against the tip of the binutils submodule in riscv-gnu-toolchain. One hunk of the patch didn't apply as the tip of the binutils submodule in riscv-gnu-toolchain is missing INSN_ALIAS on the immediate aliases of "slt" and "sltu".

I've applied this patch and the associated gcc patch <https://github.com/riscv/riscv-gcc/pull/75> to the tip of the submodules in riscv-gnu-toolchain and have successfully compiled a number of programs with `-Wa,-mstrict-operands` without any problems.

I'm currently running `check-gcc-newlib` with `RUNTESTFLAGS="--target_board=riscv-sim/-Wa,-mstrict-operands"`. I'll let you know how it goes...

[riscv-binutils-gdb-strict-operands-current.patch.txt](https://github.com/riscv/riscv-binutils-gdb/files/1059741/riscv-binutils-gdb-strict-operands-current.patch.txt)

michaeljclark:
There is no problem with gcc because it's current output and it's corrected (let's say canonical) output are both compatible with all versions of binutils. e.g. if we change gcc to emit the canonical form, there is no compatibility issue, as the documented form is already accepted by binutils.

I can't find a good rationale to not make this change when the compiler is currently emitting asm that doesn't match the spec, while binutils accepts the spec form. It would be kind of silly to not make this change as neither the gcc nor binutils changes are breaking changes.

The issue is likely to be inline asm that has been written to use the undocumented aliases.

michaeljclark:
Regarding running the test suite. I'll guess I'll have to run it twice, once without the two patches so we can check the delta. It would be nice to have a build bot to run the test suites.

aswaterman:
I don't have the resources to see this through.  If anyone wants to champion it, feel free to pick it back up.

PR #83, opened 2017-06-07
Fix issue-76 ".eh_frame_hdr refers to overlapping FDEs when linking libc"

kuanlinchentw:
Commit	4bb36f5: There is a sequential order in howto_table and riscv_reloc_map in the riscv_reloc_type_lookup design.
Commit d86acd1: To fix issue https://github.com/riscv/riscv-binutils-gdb/issues/76.


include/elf/riscv.h
```diff
@@ -87,6 +87,8 @@ START_RELOC_NUMBERS (elf_riscv_reloc_type)
   RELOC_NUMBER (R_RISCV_SET8, 54)
   RELOC_NUMBER (R_RISCV_SET16, 55)
   RELOC_NUMBER (R_RISCV_SET32, 56)
+  RELOC_NUMBER (R_RISCV_DELETE, 57)
```
aswaterman: We didn't put R_RISCV_DELETE here because it is purely internal to BFD, so shouldn't be exposed in the ELF ABI.

aswaterman: riscv_reloc_type_lookup uses a linear search, so ordering shouldn't matter...?

aswaterman:
Thanks for the fix to #76!

include/elf/riscv.h
kuanlinchentw: if "RELOC_NUMBER (R_RISCV_DELETE, 57)" is not defined, the number of R_RISCV_32_PCREL will be 57. When riscv_reloc_type_lookup searches BFD_RELOC_RISCV_32_PCREL, it return &howto_table[57]. But howo_table[57] is used by R_RISCV_DELET.

Otherwise, skip over 57 in RELOC_NUMBER. And define R_RISCV_32_PCREL  to 58 is ok, too.

palmer-dabbelt:
Thanks.  R_RISCV_DELETE needs another pass, so I won't submit this upstream until that lands as well.  Hopefully I'll have a bit of time to look at R_RISCV_DELETE soon.

include/elf/riscv.h
aswaterman: Oh, sorry, I didn't realize R_RISCV_DELETE had an entry in howto_table.

I'll figure out the best solution.

aswaterman:
https://github.com/riscv/riscv-binutils-gdb/pull/84

noamc:
I got similar linker failure for riscv64, is there a fix for 64_PCREL?

palmer-dabbelt:
@noamc Can you open a new bug with some more specifics?

noamc:
Sure,
How ?
What is the bug tracking system you use (URL please)

aswaterman:
This is the bug tracking system we prefer :) By "new issue" we just meant to start another thread: https://github.com/riscv/riscv-binutils-gdb/issues/new

PR #88, opened 2017-06-16
Allow the mixing of ".option norvc" and ".align"

palmer-dabbelt:
There's a long description in the second patch, as I'm not actually sure it's correct.

### DIFF
--- a/bfd/elfnn-riscv.c
+++ b/bfd/elfnn-riscv.c
@@ -3058,7 +3058,7 @@ _bfd_riscv_relax_tls_le (bfd *abfd,
 
 static bfd_boolean
 _bfd_riscv_relax_align (bfd *abfd, asection *sec,
-			asection *sym_sec ATTRIBUTE_UNUSED,
+			asection *sym_sec,
 			struct bfd_link_info *link_info ATTRIBUTE_UNUSED,
 			Elf_Internal_Rela *rel,
 			bfd_vma symval,
@@ -3081,7 +3081,14 @@ _bfd_riscv_relax_align (bfd *abfd, asection *sec,
 
   /* Make sure there are enough NOPs to actually achieve the alignment.  */
   if (rel->r_addend < nop_bytes)
-    return FALSE;
+    {
+      (*_bfd_error_handler)
+	(_("%B(%A+0x%lx): %d bytes required for alignment"
+	   "to %d-byte boundary, but only %d present"),
+	   abfd, sym_sec, rel->r_offset, nop_bytes, alignment, rel->r_addend);
+      bfd_set_error (bfd_error_bad_value);
+      return FALSE;
+    }
 
   /* Delete the reloc.  */
   rel->r_info = ELFNN_R_INFO (0, R_RISCV_NONE);
--- a/bfd/elfnn-riscv.c
+++ b/bfd/elfnn-riscv.c
@@ -2704,10 +2704,142 @@ riscv_relax_delete_bytes (bfd *abfd, asection *sec, bfd_vma addr, size_t count)
   return TRUE;
 }
 
+/* A second format for recording PC-relative hi relocations.  This stores the
+   information required to relax them to GP-relative addresses.  */
+
+typedef struct riscv_pcgp_hi_reloc riscv_pcgp_hi_reloc;
+struct riscv_pcgp_hi_reloc
+{
+  bfd_vma hi_sec_off;
+  bfd_vma hi_addend;
+  bfd_vma hi_addr;
+  unsigned hi_sym;
+  asection *sym_sec;
+  riscv_pcgp_hi_reloc *next;
+};
+
+typedef struct riscv_pcgp_lo_reloc riscv_pcgp_lo_reloc;
+struct riscv_pcgp_lo_reloc
+{
+  bfd_vma hi_sec_off;
+  riscv_pcgp_lo_reloc *next;
+};
+
+typedef struct
+{
+  riscv_pcgp_hi_reloc *hi;
+  riscv_pcgp_lo_reloc *lo;
+} riscv_pcgp_relocs;
+
+static bfd_boolean
+riscv_init_pcgp_relocs (riscv_pcgp_relocs *p)
+{
+  p->hi = NULL;
+  p->lo = NULL;
+  return TRUE;
+}
+
+static bfd_boolean
+riscv_free_pcgp_relocs (riscv_pcgp_relocs *p, bfd *abfd, asection *sec)
+{
+  for (riscv_pcgp_hi_reloc *c = p->hi; c != NULL;)
+    {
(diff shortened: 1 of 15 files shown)


bfd/elfnn-riscv.c
```diff
@@ -3081,7 +3081,13 @@ _bfd_riscv_relax_align (bfd *abfd, asection *sec,
 
   /* Make sure there are enough NOPs to actually achieve the alignment.  */
   if (rel->r_addend < nop_bytes)
-    return FALSE;
+    {
+      (*_bfd_error_handler)
+	(_("%B(%A+0x%lx): %d bytes required for alignment to %d-byte boundry, but only %d present"),
```
aswaterman: boundary is misspelled

palmer-dabbelt:
Here's another pathological test case:

````
.text
.global _prog_start
_prog_start:
  la x2, data
  j code

data:
.ascii "abcdefg"
.align 2

code:
  add x0, x1, x1

.align 4
  mv x1, x1
  mv x2, x2
.align 2
rvc_enabled:
.option norvc
rvc_disabled:
.align 3
align_3:
  add x0, x4, x4
````

````
0000000080000000 <_prog_start>:
    80000000:   00000097                auipc   ra,0x0
    80000004:   00a08093                addi    ra,ra,10 # 8000000a <data>
    80000008:   a031                    j       80000014 <code>

000000008000000a <data>:
    8000000a:   6261                    lui     tp,0x18
    8000000c:   66656463                bltu    a0,t1,80000674 <align_3+0x64c>
    80000010:   00010067                jr      sp

0000000080000014 <code>:
    80000014:   00108033                add     zero,ra,ra
    80000018:   00000013                nop
    8000001c:   00000013                nop
    80000020:   8086                    mv      ra,ra
    80000022:   810a                    mv      sp,sp

0000000080000024 <rvc_disabled>:
    80000024:   00000013                nop

0000000080000028 <align_3>:
    80000028:   00420033                add     zero,tp,tp
````

**Comment by PkmX** [comment]
With this pull request, `.p2align n` always emits `2^n-1` bytes of NOP, so this forces the linker to relax away the extra padding. This complicates linkers without support for relaxation (see discussion with lld [here](http://lists.llvm.org/pipermail/llvm-dev/2017-July/115310.html)).

Would it be feasible to add a *command-line option* to `as` that assumes no relaxation? I'm aware of `.option norelax` but it seems to be broken with this patch:

```asm
.option norelax
_start:
    mv  x1, x1s
    .align 2
    mv  x2, x2
```

It generates two invalid instructions and no `R_RISCV_ALIGN` reloc:

```
00000000 <_start>:
   0:	8086                	mv	ra,ra
   2:	0000                	unimp
   4:	0000                	unimp
   6:	0001                	nop
   8:	810a                	mv	sp,sp
```

aswaterman:
IMO we don't need to support linkers that don't support relaxation; it is a
fundamental expectation of the RISC-V linking model. Better to improve lld.

sorear:
How does alignment work on other architectures?

This will also be a problem for the Go linker, which does not support relocations that change symbol addresses at present.

brucehoult:
@aswaterman I think it's a useful property to make it *possible* to run the code with the linker only performing relocations, but not resulting simplifications. Of course it is desirable to also reduce the code size and instruction count where possible.

It seems somehow ... perverse ... to always e.g. emit 3 bytes of padding when four byte alignment is requested. Why not 4, at least? Sure, it's not strictly necessary in that sometimes the linker will delete all four, but it's somehow prettier.

Also, why not ensure that the requested alignment is already present in the object file, so that it is never *logically* necessary for the linker to change the code size.

How about this?

```c
/*
  RISC-V align.

  Goals:
  1) align to the desired amount in the generated object file, so that it will work with
     pure non-size-changing relocations only.
  2) provide enough extra padding that alignment can be maintained in the face of
     code size changes, by only making the padding smaller, never bigger.
*/

void emit_padding(uintptr_t sz);

uintptr_t align(uintptr_t oldPos, int logAligment){
    uintptr_t alignment = 1 << logAligment, mask = alignment - 1;
    uintptr_t newPos = (oldPos + mask) & mask;
    if ((newPos - oldPos) < mask) newPos += alignment;
    emit_padding(newPos - oldPos);
    return newPos;
}
```

espindola:
@aswaterman We definitely want to support relaxations is lld. That is not the reason why we object to them being required.

By making all optimizations optional, we can have  a really fast link for use in developer builds. That is implemented in lld with the -O0 option. In this mode we don't even merge strings, it would be silly to have to do relaxations.

The way risc-v is going seems better than other architectures. ARM and Mips use an optimistic compiler/assembler and the linker *has* to fix that up. Worse still, multiple iterations might be required as adding thunks can cause other references to go out of range.

It seems that RISC-V uses a pessimistic compiler/assembler and the linker *can* optimize that. If distances never increase, each iteration ends with a valid layout and we don't have to iterate multiple times unless we want to check for extra optimization opportunities.

To guarantee that distances never increase across a branch, it has to be of at least 2^n-1 bytes. As @brucehoult  suggests, the assembler can output padding that is in the range [2^n-1, 2^n + n - 1). With that we have both properties:

* Sections can be just concatenated (relaxations are optional).
* Distances will never increase (we can do any number of relaxation rounds)

aswaterman:
@espindola Fast link seems like a good justification to me.  And I agree that the compiler/assembler being pessimistic makes that quite a bit easier.

brucehoult:
@aswaterman Great!

I just realised there's a simpler way to formulate it, without any conditional code. Add 2^n-1 as originally proposed here, but then round to the next 2^n by adding 2^n-1 again and masking:

``uintptr_t newPos = (oldPos + mask + mask) & ~mask;``

palmer-dabbelt:
The reason I didn't round to 2^n is because if this code in BFD: <https://github.com/riscv/riscv-binutils-gdb/blob/riscv-next/bfd/elfnn-riscv.c#L3079>

````
  while (alignment <= rel->r_addend)
    alignment *= 2;
````

I think it's illegal to change that, as it'd be an ABI-incompatible change.  We could add another alignment relocation with better semantics.

palmer-dabbelt:
I'm marking all these as WIP and merging this (manually).  For the case where you don't want linker relaxations you should get the right thing by adding ".option norelax" to the top of every assembly file.  I don't have time to test that right now, but I'll try to remember to do that before upstreaming it.

If you've got time to test this and it doesn't work then please file a bug.

PkmX:
> For the case where you don't want linker relaxations you should get the right thing by adding ".option norelax" to the top of every assembly file.

I think it would be useful to have a `--no-relax` command line argument for the assembler to do that, so users can simply invoke gcc with `-Wa,--no-relax`.

palmer-dabbelt:
This is actually something that should live in GCC.  Can you file a bug to add a "-mno-relax" option to GCC so I don't forget?  Thanks.

PR #91, opened 2017-06-28
Fix TLS copy relocs

aswaterman:
The dynrelro introduction wasn't implemented quite right for RISC-V, as it didn't consider TLS copy relocs.

This fixes tls-tst10 in the glibc test suite.


palmer-dabbelt:
This didn't add any new failures, so I sent it upstream and merged it to riscv-next.  GitHub is confused, so I'm closing the PR.

PR #92, opened 2017-07-18
Add riscv_elf_ignore_reloc to pass gcc_cv_as_cfi_advance_working.

kuanlinchentw:
This patch is for GCC that can generate human readable cfi directive as @kito-cheng mentioned in [riscv-gcc/issues/81](https://github.com/riscv/riscv-gcc/issues/81#issuecomment-315976906).

### DIFF
--- a/bfd/elfxx-riscv.c
+++ b/bfd/elfxx-riscv.c
@@ -32,6 +32,9 @@
 
 #define MINUS_ONE ((bfd_vma)0 - 1)
 
+static bfd_reloc_status_type riscv_elf_ignore_reloc
+(bfd *, arelent *, asymbol *, void *, asection *, bfd *, char **);
+
 /* The relocation table used for SHT_RELA sections.  */
 
 static reloc_howto_type howto_table[] =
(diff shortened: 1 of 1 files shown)


kito-cheng:
How gcc detect this : https://github.com/riscv/riscv-gcc/blob/riscv-next/gcc/configure.ac#L2804

aswaterman:
Thanks @kuanlinchentw!  But I wonder, is the test in gcc/configure.ac broken?  (If you guys think this is the best fix, I approve)

kito-cheng:
No, gcc is using right way for detect this, and we think it's best fix now :)
btw, this fix already tested ~6 month according our internal git log :P

kuanlinchentw:
@palmer-dabbelt I found [issue-248](https://github.com/riscv/riscv-gnu-toolchain/issues/248) is similar problem with this pull request.  Please close the pull request without merging it into the upstream branch.
I will create another pull request to fix both issue at once. Thanks.

palmer-dabbelt:
Thanks!

PR #98, opened 2017-07-25
Add special_function for ADD and SUB relocations.

kuanlinchentw:
For [issue-248](https://github.com/riscv/riscv-gnu-toolchain/issues/248) to print the correct line number.
For [issue-81](https://github.com/riscv/riscv-gcc/issues/81#issuecomment-315976906) to make GCC generate the human readable cfi directive.


atraber:
I think there is an alternative for this special reloc function, because one could just use the generic reloc function for at least R_RISCV_ADD8, R_RISCV_ADD16, R_RISCV_ADD32, R_RISCV_ADD64, R_RISCV_SUB16 and R_RISCV_SUB32.

When you set the src_mask in the HOWTO to MINUS_ONE, the addition will be done also with bfd_elf_generic_reloc. For subtraction one can use the special sizes -1 and -2. So this works at least for SUB16 and SUB32. Checkout https://github.com/riscv/riscv-binutils-gdb/blob/riscv-next/bfd/reloc.c#L891

kuanlinchentw:
I think R_RISCV_ADDXX relocations can use generic reloc function if they do not use the value in the section content.  (As far as I know, these relocations seems do not use so far.)
But SUB16 and SUB32 can't use bfd_elf_generic_reloc because their HOWTO src_mask cannot be set to MINUS_ONE. In [src_mask limit](https://github.com/riscv/riscv-binutils-gdb/blob/riscv-next/bfd/bfd-in2.h#L2599)->"If relocations do have an addend in the reloc, eg. ELF USE_RELA, this field should be zero."

palmer-dabbelt:
This was merged here /nettmp/netapp1a/jenkins/archived-builds/nightly/federation-clean-nightly-param-g5soc-test-nightly/4015/builds/g5soc_sim/sim/sim/hart1/rv64mi-ma_addr/program.elf.err

PR #100, opened 2017-07-26
Don't treat Q extension as special case

aswaterman:
@kito-cheng on the Binutils mailing list, Klaus Pedersen pointed out that the Q extension not be a special case in ISA-string parsing.  I believe he's correct; this patch simplifies the code and enforces the ISA-string spec (Q must come after D and before C).


aswaterman:
this patch isn't quite right - fixing

aswaterman:
ok, I think it's right this time.

This patch fixes a segfault for ```-march=rv64imaqc```.

PR #101, opened 2017-08-06
RISC-V: Mark various compressed aliases as aliases

palmer-dabbelt:
This fixes "-M noaliases" disassembly for these instructions.

opcodes/ChangeLog

2017-08-01  Palmer Dabbelt  <palmer@dabbelt.com>

        * riscv-opc.c (riscv_opcodes): Mark "c.jr", "c.j", "c.mv",
        "c.lui" and "c.li" as aliases.


palmer-dabbelt:
@aswaterman these are actually aliases, right?

aswaterman:
Just ```c.nop```, which is an alias for ```c.addi x0, 0```.  The others are all real instructions.

palmer-dabbelt:
@aswaterman better?

PR #102, opened 2017-08-15
Check ISA extension prerequisites

aswaterman:
2017-08-15  Andrew Waterman  <andrew@sifive.com>

	* config/tc-riscv.c (riscv_set_arch): Check that D and Q extensions
	meet their prerequisites.

### DIFF
--- a/gas/config/tc-riscv.c
+++ b/gas/config/tc-riscv.c
@@ -215,6 +215,15 @@ riscv_set_arch (const char *s)
 	as_fatal ("-march=%s: unsupported ISA subset `%c'", s, *p);
     }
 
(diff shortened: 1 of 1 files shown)


aswaterman:
rebased

kito-cheng:
btw, Q ext for GCC I've already done before, but just no simulator can run the testsuite yet.

aswaterman:
I'll add RV64Q support to Spike soon.  (Spike can run some of the testsuite using newlib and pk, but qemu support is more important.)

aswaterman:
I think @kito-cheng included this patch as part of #103, so I'll close it

PR #103, opened 2017-08-24
Support RV32E

kito-cheng:
Open a new PR for RV32E :P

### DIFF
--- a/gas/config/tc-riscv.c
+++ b/gas/config/tc-riscv.c
@@ -215,6 +215,15 @@ riscv_set_arch (const char *s)
 	as_fatal ("-march=%s: unsupported ISA subset `%c'", s, *p);
     }
 
+  if (riscv_subset_supports ("d") && !riscv_subset_supports ("f"))
+    as_fatal ("-march=%s: `d' extension requires `f' extension", s);
+
+  if (riscv_subset_supports ("q") && !riscv_subset_supports ("d"))
+    as_fatal ("-march=%s: `q' extension requires `d' extension", s);
+
+  if (riscv_subset_supports ("q") && xlen < 64)
+    as_fatal ("-march=%s: rv32 does not support the `q' extension", s);
+
   free (extension);
 }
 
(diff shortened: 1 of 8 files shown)


gas/config/tc-riscv.c
```diff
@@ -171,6 +182,13 @@ riscv_set_arch (const char *s)
       case 'i':
 	break;
 
+      case 'e':
+	p++;
+	riscv_add_subset ("e");
+	riscv_add_subset ("i");
+	riscv_set_rve (TRUE);
```
aswaterman: I think we want to add

    if (xlen > 32)
      as_fatal ("-march=%s: rv%de is not a valid base ISA", s, xlen);

aswaterman:
I think we should merge https://github.com/riscv/riscv-binutils-gdb/pull/102 first, then add a similar check to enforce prerequisites:

    if (riscv_subset_supports ("e") && riscv_subset_supports ("f"))
      as_fatal ("-march=%s: rv32e does not support the `f' extension", s);

aswaterman:
Also, note that RV32E Is only version 1.9, so there may be some changes before it is frozen.  In particular, the C extension might be slightly different for RVE (e.g., ```c.lbu``` and ```c.lh``` might replace ```c.flw``` and ```c.fld```).

I think we can merge this patch anyway--just a warning.

kito-cheng:
> Also, note that RV32E Is only version 1.9, so there may be some changes before it is frozen. In particular, the C extension might be slightly different for RVE (e.g., c.lbu and c.lh might replace c.flw and c.fld).

Ya, we are considering to recycle those encoding space too.

gas/config/tc-riscv.c
kito-cheng: Done.

kito-cheng:
Rebase to #102,

palmer-dabbelt:
I'll submit this upstream in a bit, everything's a bit of a mess right now...

kito-cheng:
I guess maybe I can back to support the gcc part next month, we are mess in several meaning too...:P

PR #105, opened 2017-09-24
Only relax to C.LUI when imm != 0 and rd != 0/2

aswaterman:


palmer-dabbelt:
I broke this because of a rebase.  This is on riscv-next: https://github.com/riscv/riscv-binutils-gdb/commit/96f3fcdc4bbab54d58cd09eb0ef66d2d0ca72ada

PR #111, opened 2017-11-01
Use ABI names for registers.

timsifive:
This should cause the machine interface register list to list the ABI
names, which in turn will cause them to show up that way in Eclipse.
(Eclipse can apparently not give the registers useful names itself.)

This fixes https://github.com/riscv/riscv-openocd/issues/93

### DIFF
--- a/gdb/riscv-tdep.c
+++ b/gdb/riscv-tdep.c
@@ -80,7 +80,7 @@ struct riscv_frame_cache
 static const char * const riscv_gdb_reg_names[RISCV_LAST_FP_REGNUM + 1] =
 {
   "zero", "ra", "sp", "gp", "tp", "t0", "t1", "t2",
-  "fp", "s1", "a0", "a1", "a2", "a3", "a4", "a5",
+  "s0", "s1", "a0", "a1", "a2", "a3", "a4", "a5",
   "a6", "a7", "s2", "s3", "s4", "s5", "s6", "s7",
   "s8", "s9", "s10", "s11", "t3", "t4", "t5", "t6",
   "pc",
@@ -107,7 +107,7 @@ static const struct register_alias riscv_register_aliases[] =
   { "x6", 6 },
   { "x7", 7 },
   { "x8", 8 },
-  { "s0", 8 },
+  { "fp", 8 },
   { "x9", 9 },
   { "x10", 10 },
   { "x11", 11 },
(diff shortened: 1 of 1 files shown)


gdb/riscv-tdep.c
```diff
@@ -75,17 +75,19 @@ struct riscv_frame_cache
...
+/* Use the ABI names here because Eclipse can't display names besides what we
+ * declare here, and users want to see the ABI names in any case. */
 static const char * const riscv_gdb_reg_names[RISCV_LAST_FP_REGNUM + 1] =
 {
-  "x0",  "x1",  "x2",  "x3",  "x4",  "x5",  "x6",  "x7",
-  "x8",  "x9",  "x10", "x11", "x12", "x13", "x14", "x15",
-  "x16", "x17", "x18", "x19", "x20", "x21", "x22", "x23",
-  "x24", "x25", "x26", "x27", "x28", "x29", "x30", "x31",
+  "zero", "ra", "sp", "gp", "tp", "t0", "t1", "t2",
+  "fp", "s1", "a0", "a1", "a2", "a3", "a4", "a5",
```
ilg-ul: I checked the disassembled code and, at least in my environment, it uses `s0`, not `fp`:

```
40400196:	1141                	addi	sp,sp,-16
40400198:	c622                	sw	s0,12(sp)
4040019a:	0800                	addi	s0,sp,16
```

aswaterman: Indeed, s0 is the preferred alias for x8.

ilg-ul:
I checked this patch and except the fp/s0 detail, it seems ok, the API register names are shown in the Registers view, see image.
screen shot 2017-11-02 at 10 08 51

timsifive:
Thanks for reviewing.

TM1234:
This doesn't do anything for CSR names.
They're still all CSRn (0 <= n <= 4095) rather than using, say, draft priv abi spec v0.10 names.
So I'm not sure it's correct to say that this mod fixes https://github.com/riscv/riscv-openocd/issues/93.
And the correct solution is still to use the gdb/mi XML mechanism but I guess that nobody has had time yet to understand this fully and propose a solution that uses it (including me!)?

ilg-ul:
I renamed https://github.com/riscv/riscv-openocd/issues/93 to explicitly refer to Core Registers.

PR #112, opened 2017-11-01
We don't implement any pseudoregisters.

timsifive:
I assume this is just some old cut-and-paste code. Without this change,
the machine interface reports a large number of unnamed registers at the
end of the real list in response to `data-list-register-names`.

### DIFF
--- a/gdb/riscv-tdep.c
+++ b/gdb/riscv-tdep.c
@@ -456,28 +456,6 @@ riscv_return_value (struct gdbarch  *gdbarch,
   return RETURN_VALUE_REGISTER_CONVENTION;
 }
 
-/* Implement the pseudo_register_read gdbarch method.  */
-
-static enum register_status
-riscv_pseudo_register_read (struct gdbarch *gdbarch,
-			    struct regcache *regcache,
-			    int regnum,
-			    gdb_byte *buf)
-{
-  return regcache_raw_read (regcache, regnum, buf);
-}
-
-/* Implement the pseudo_register_write gdbarch method.  */
-
-static void
-riscv_pseudo_register_write (struct gdbarch *gdbarch,
-			     struct regcache *regcache,
-			     int cookednum,
-			     const gdb_byte *buf)
-{
-  regcache_raw_write (regcache, cookednum, buf);
-}
-
 /* Implement the register_type gdbarch method.  */
 
 static struct type *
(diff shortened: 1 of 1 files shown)


ilg-ul:
I checked this patch too and it reduced the list of registers, it now reads

```
1-data-list-register-names
1^done,register-names=["zero","ra","sp","gp","tp","t0","t1","t2","fp","s1","a0","a1","a2","a3","a4","a5","a6","a7","s2","s3","s4","s5","s6","s7","s8","s9","s10","s11","t3","t4","t5","t6","pc","ft0","ft1","ft2","ft3","ft4","ft5","ft6","ft7","fs0","fs1","fa0","fa1","fa2","fa3","fa4","fa5","fa6","fa7","fs2","fs3","fs4","fs5","fs6","fs7","fs8","fs9","fs10","fs11","ft8","ft9","ft10","ft11","csr0","csr1","csr2","csr3","csr4","csr5","csr6","csr7","csr8","csr9","csr10","csr11","csr12","csr13","csr14",...,"csr4093","csr4094","csr4095","priv"]
(gdb) 
```

(I manually cut a big part of it to make it more readable)

I think this patch should be merged. 

Unfortunately (as expected otherwise) this is not enough to prevent Eclipse hanging in the Memory View tab... keep digging...

ilg-ul:
not directly related to the subject, but please note that while working with `riscv-tdep.c` I noticed several lines starting with TABs, while most of the file is indented with spaces.

I don't know what editor you are using, but I would change the setting to convert TABs to spaces.

as it is now, when seen with an editor that has a small TAB size, the text looks poorly formatted.

gdb/riscv-tdep.c
```diff
@@ -1286,10 +1264,7 @@ riscv_gdbarch_init (struct gdbarch_info info,
   set_gdbarch_sw_breakpoint_from_kind (gdbarch, riscv_sw_breakpoint_from_kind);
 
   /* Register architecture.  */
-  set_gdbarch_pseudo_register_read (gdbarch, riscv_pseudo_register_read);
-  set_gdbarch_pseudo_register_write (gdbarch, riscv_pseudo_register_write);
   set_gdbarch_num_regs (gdbarch, RISCV_NUM_REGS);
-  set_gdbarch_num_pseudo_regs (gdbarch, RISCV_NUM_REGS);
```
ilg-ul: I did not check the default value for this field, but a defensive approach would be to explicitly initialise it at 0.

```c
// No pseudoregisters.
set_gdbarch_num_pseudo_regs (gdbarch, 0);
```

timsifive:
All new gdbarch structures are explicitly zeroed in gdbarch_alloc().

PR #113, opened 2017-11-11
RISC-V: Enforce 4-byte alignment before '.option norvc'

palmer-dabbelt:
RISC-V systems without the C extension mandate 4-byte alignment.  Users
keep forgetting about this and filing bugs, so we're just going to go
ahead and enforce the alignment as part of '.option norvc'.

gas/ChangeLog

2017-11-08  Palmer Dabbelt  <palmer@sifive.com>

        * config/tc-riscv.c (s_riscv_option): Enforce 4-byte alignment
        on '.option norvc'.


palmer-dabbelt:
I haven't even compiled this yet.

palmer-dabbelt:
Oh, actually I just saw an email from Andrew.  We're going to improve the quality of the error message instead of adding implicit behavior.

aswaterman:
cool

PR #115, opened 2017-11-15
Add support for servers that send an XML target description

timsifive:
This is the first step towards having OpenOCD expose only those registers that actually exist.

This change works with current OpenOCD, as well as my dev version which sends a complete list of registers. Tested against spike, as well as 3 hardware implementations.


ilg-ul:
interesting. does the default makefile compile `gdb/riscv-tdep.c` as C++? I expected it to be `.cpp`, or `.cc`, etc.

jim-wilson:
GDB was originally C code, switched to a C++ compiler last year, and is now converting one type at a time to C++.  SInce it was C code, most of the files have .c endings, and this was not changed because renaming hundreds of files is a hassle.  GCC has the same issue, most files are .c, but they are compiled by a C++ compiler.
https://sourceware.org/ml/gdb-patches/2016-09/msg00029.html

ilg-ul:
Thank you, Jim.

timsifive:
Can I get a review on this, or should I just merge it?

ilg-ul:
you already have an openocd that sends the xml, and its content is properly passed to gdb? 

what is the logic, if the xml is received, it completely replaces the static definitions in gdb?

timsifive:
>
> you already have an openocd that sends the xml, and its content is
> properly passed to gdb?
>
Yes. It's still work-in-progress, but currently it does not send FPRs when
the hart doesn't contain an FPU.

> what is the logic, if the xml is received, it completely replaces the
> static definitions in gdb?
>
Effectively, yes.

ilg-ul:
did you check how the same functionality is implemented for ARM GDB?

timsifive:
I looked at a few other gdb targets. I don't recall if ARM was one of them.

ilg-ul:
I'm a bit surprised by the extent of the changes. I expected the code that processes the xml to be common to all architectures, and require minimal support from each specific architecture.

if I'm wrong, and each architecture has a different way of managing these xmls, I think the ARM code should be the main inspiration, if for no other reason, at least to help SEGGER, they already do this for ARM in their code, and it would be easier to replicate the same solution for RISC-V too.

gdb/gdbarch.c
```diff
@@ -3569,7 +3574,8 @@ gdbarch_register_reggroup_p (struct gdbarch *gdbarch, int regnum, struct reggrou
   gdb_assert (gdbarch != NULL);
   gdb_assert (gdbarch->register_reggroup_p != NULL);
   if (gdbarch_debug >= 2)
-    fprintf_unfiltered (gdb_stdlog, "gdbarch_register_reggroup_p called\n");
+    fprintf_unfiltered (gdb_stdlog, "gdbarch_register_reggroup_p(%d) called\n",
+                        regnum);
```
jim-wilson: Changes to gdbarch.c will have to be submitted separately from the riscv port.  You are printing regnum here, but not the register group name, which might also be useful.

gdb/riscv-tdep.c
```diff
@@ -75,100 +76,101 @@ struct riscv_frame_cache
...
+  {RISCV_S8_REGNUM, {"s8", "x24"}, true, true, "org.gnu.gdb.riscv.cpu", general_reggroup},
+  {RISCV_S9_REGNUM, {"s9", "x25"}, true, true, "org.gnu.gdb.riscv.cpu", general_reggroup},
+  {RISCV_S10_REGNUM, {"s10", "x26"}, true, true, "org.gnu.gdb.riscv.cpu", general_reggroup},
+  {RISCV_S11_REGNUM, {"s11", "x27"}, true, true, "org.gnu.gdb.riscv.cpu", general_reggroup},
+  {RISCV_T3_REGNUM, {"t3", "x28"}, true, true, "org.gnu.gdb.riscv.cpu", general_reggroup},
+  {RISCV_T4_REGNUM, {"t4", "x29"}, true, true, "org.gnu.gdb.riscv.cpu", general_reggroup},
+  {RISCV_T5_REGNUM, {"t5", "x30"}, true, true, "org.gnu.gdb.riscv.cpu", general_reggroup},
+  {RISCV_T6_REGNUM, {"t6", "x31"}, true, true, "org.gnu.gdb.riscv.cpu", general_reggroup},
+
+  {RISCV_PC_REGNUM, {"pc"}, true, true, "org.gnu.gdb.riscv.cpu", general_reggroup},
```
jim-wilson: GCC puts the FP registers immediately after the general registers, which means FP register numbers won't match between gcc and gdb unless they are being remapped someplace else.

gdb/riscv-tdep.c
```diff
@@ -1301,42 +1241,139 @@ riscv_gdbarch_init (struct gdbarch_info info,
...
+#include "opcode/riscv-opc.h"
+#undef DECLARE_CSR
+  };
+
+  static bool reg_info_built = false;
+  if (!reg_info_built)
+    {
+      for (unsigned i = 0; i < sizeof(named_csr) / sizeof(*named_csr); i++)
+        {
+          char *generic_name = (char*) malloc(8);
```
jim-wilson: This fails if there are more than 9999 CSRs.  I would suggest checking for that and aborting to avoid writing past the end of the malloced area.

gdb/riscv-tdep.c
```diff
@@ -1301,42 +1241,139 @@ riscv_gdbarch_init (struct gdbarch_info info,
...
+      if (use_tdesc_registers)
+        {
+          // This number apparently must be at least as large as the number of
+          // registers in the target description. (See assertion in
+          // tdesc_use_registers().) But that seems to defeat part of the point
+          // of the target description, so I might be missing something.
+          set_gdbarch_num_regs (gdbarch, RISCV_NUM_REGS);
+          set_gdbarch_sp_regnum (gdbarch, RISCV_SP_REGNUM);
+          set_gdbarch_pc_regnum (gdbarch, RISCV_PC_REGNUM);
+          set_gdbarch_ps_regnum (gdbarch, RISCV_FP_REGNUM);
```
jim-wilson: Setting the processor status register to point at the frame pointer register looks odd, but it was that way before your patch.

gdb/gdbarch.c
timsifive: Yeah. If they don't like these changes we can just axe them. But I think they help in debugging.
I'll print reggroup as a pointer. Not sure how helpful that is, but it doesn't hurt and it's easy.

gdb/riscv-tdep.c
timsifive: That's interesting. I don't know how that mapping works, but clearly it must. The actual numbers used aren't actually changed in this changed. RISCV_PC_REGNUM has had that value since before I started working on it.

gdb/riscv-tdep.c
timsifive: Added an assert.

gdb/riscv-tdep.c
timsifive: The actual frame examining code is voodoo to me, so I just try not to touch it.

gdb/riscv-tdep.c
jim-wilson: Maybe no one has tested gdb on a cpu that has hardware floating point?

timsifive: I test against spike with FPU almost every time I push a debug-related change. Floating support definitely works.

timsifive:
@jim-wilson Thanks for reviewing.

jim-wilson:
I just noticed that this was a pull request for riscv-binutils-2.29.  Shouldn't this patch also be on the riscv-next branch?  Otherwise, we appear to have two divergent gdb ports, one on the riscv-next branch, and one on the riscv-binutils-2.29 branch.  I think patches should go on riscv-next first, and then cherry picked onto riscv-binutils-2.29 if needed there.

timsifive:
That's an excellent question. I'm under the impression that riscv-binutils-2.29 is "main" in this repo because it's the one that shows up when you visit https://github.com/riscv/riscv-binutils-gdb, but I may be wrong.

@palmer-dabbelt @aswaterman  Is this documented anywhere?

palmer-dabbelt:
I think I just commented on another issue, but let me write up an email and try to get this cleaned up a bit.

PR #116, opened 2017-11-15
GDB sim

kito-cheng:
This PR contain E, C, F and D extension and several  gdb sim fix.

### DIFF
--- a/sim/riscv/interp.c
+++ b/sim/riscv/interp.c
@@ -27,6 +27,10 @@
 
 #include "config.h"
 
+#include "libiberty.h"
+#include "bfd.h"
+#include "elf-bfd.h"
+
 #include "sim-main.h"
 #include "sim-options.h"
 
@@ -154,6 +158,8 @@ sim_create_inferior (SIM_DESC sd, struct bfd *abfd,
 {
   SIM_CPU *cpu = STATE_CPU (sd, 0);
   sim_cia addr;
+  Elf_Internal_Phdr *phdr;
+  int i, phnum;
 
   /* Set the PC.  */
   if (abfd != NULL)
@@ -161,6 +167,14 @@ sim_create_inferior (SIM_DESC sd, struct bfd *abfd,
   else
     addr = 0;
   sim_pc_set (cpu, addr);
+  phdr = elf_tdata (abfd)->phdr;
+  phnum = elf_elfheader (abfd)->e_phnum;
+
+  for (i = 0; i < phnum; i++)
+    {
+      if (phdr[i].p_paddr + phdr[i].p_memsz > cpu->endbrk)
+	cpu->endbrk = phdr[i].p_paddr + phdr[i].p_memsz;
+    }
 
   /* Standalone mode (i.e. `run`) will take care of the argv for us in
      sim_open() -> sim_parse_args().  But in debug mode (i.e. 'target sim'
--- a/sim/riscv/interp.c
+++ b/sim/riscv/interp.c
@@ -148,6 +148,27 @@ sim_open (SIM_OPEN_KIND kind, host_callback *callback,
   return sd;
 }
 
+static bfd_vma
+riscv_get_symbol (SIM_DESC sd, const char *sym)
+{
+  long symcount = STATE_PROG_SYMS_COUNT (sd);
+  asymbol **symtab = STATE_PROG_SYMS (sd);
+  int i;
+
+  for (i = 0;i < symcount; ++i)
+    {
+      if (strcmp (sym, bfd_asymbol_name (symtab[i])) == 0)
+	{
+	  bfd_vma sa;
+	  sa = bfd_asymbol_value (symtab[i]);
+	  return sa;
+	}
+    }
+
+  /* Symbol not found.  */
+  return 0;
+}
+
 /* Prepare to run a program that has already been loaded into memory.
 
    Usually you do not need to change things here.  */
@@ -170,10 +191,18 @@ sim_create_inferior (SIM_DESC sd, struct bfd *abfd,
   phdr = elf_tdata (abfd)->phdr;
   phnum = elf_elfheader (abfd)->e_phnum;
 
-  for (i = 0; i < phnum; i++)
+  /* Try to find _end symbol, and set it to the end of brk.  */
+  trace_load_symbols (sd);
+  cpu->endbrk = riscv_get_symbol (sd, "_end");
+
+  /* If not found, set end of brk to end of all section.  */
+  if (cpu->endbrk == 0)
     {
-      if (phdr[i].p_paddr + phdr[i].p_memsz > cpu->endbrk)
-	cpu->endbrk = phdr[i].p_paddr + phdr[i].p_memsz;
+      for (i = 0; i < phnum; i++)
+	{
+	  if (phdr[i].p_paddr + phdr[i].p_memsz > cpu->endbrk)
+	    cpu->endbrk = phdr[i].p_paddr + phdr[i].p_memsz;
+	}
     }
 
   /* Standalone mode (i.e. `run`) will take care of the argv for us in
--- a/sim/riscv/sim-main.c
+++ b/sim/riscv/sim-main.c
@@ -1159,6 +1159,8 @@ void initialize_env (SIM_DESC sd, const char * const *argv,
   sp = sp_flat - ((argc + 1 + envc + 1) * sizeof (address_word));
   /* Then the argc.  */
   sp -= sizeof (unsigned_word);
+  /* Synchronize sp alignment with GCC's STACK_BOUNDARY.  */
+  sp = ALIGN_16 (sp - 15);
 
   /* Set up the regs the libgloss crt0 expects.  */
   cpu->a0 = argc;
--- a/sim/riscv/sim-main.c
+++ b/sim/riscv/sim-main.c
@@ -25,6 +25,7 @@
 
 #include <inttypes.h>
 #include <time.h>
+#include <unistd.h>
 
 #include "sim-main.h"
 #include "sim-fpu.h"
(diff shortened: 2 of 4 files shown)


sim/riscv/sim-main.c
```diff
@@ -1451,7 +1451,7 @@ execute_i (SIM_CPU *cpu, unsigned_word iw, const struct riscv_opcode *op)
       store_rd (cpu, rd, EXTEND32 (ashiftrt ((signed32)cpu->regs[rs1], cpu->regs[rs2] & 0x1f)));
       break;
     case MATCH_SRAI:
-      TRACE_INSN (cpu, "srai %s, %s, %"PRIiTW";  // %s = %s >>> %#"PRIxTW,
+      TRACE_INSN (cpu, "srai %s, %s, %" PRIiTW ";  // %s = %s >>> %#"PRIxTW,
```
jim-wilson: You put spaces around the first PRIiTW but not the second one.  This is a minor nit.

sim/riscv/sim-main.c
```diff
@@ -1467,23 +1467,30 @@ execute_i (SIM_CPU *cpu, unsigned_word iw, const struct riscv_opcode *op)
...
 		  rd_name, rs1_name, shamt_imm, rd_name, rs1_name, shamt_imm);
       RISCV_ASSERT_RV64 (cpu, "insn: %s", op->name);
-      store_rd (cpu, rd, EXTEND32 (ashiftrt ((signed32)cpu->regs[rs1], shamt_imm)));
+      store_rd (cpu, rd,
+		EXTEND32 (ashiftrt ((signed32)cpu->regs[rs1], shamt_imm)));
       break;
     case MATCH_SLT:
-      TRACE_INSN (cpu, "slt");
-      store_rd (cpu, rd, !!((signed_word)cpu->regs[rs1] < (signed_word)cpu->regs[rs2]));
+      TRACE_INSN (cpu, "slt %s %s %s", rd_name, rs1_name, rs2_name);
```
jim-wilson: I would expect commas between the instruction operands.  This is a minor nit.

sim/riscv/sim-main.c
```diff
@@ -1286,13 +1286,19 @@ execute_i (SIM_CPU *cpu, unsigned_word iw, const struct riscv_opcode *op)
   unsigned_word sb_imm = EXTRACT_SBTYPE_IMM (iw);
   unsigned_word shamt_imm = ((iw >> OP_SH_SHAMT) & OP_MASK_SHAMT);
   unsigned_word tmp;
+  unsigned_word sys_id;
+  int eh_rve_p = cpu->elf_flags & 0x8;
```
jim-wilson: Should use an EF_* value from include/elf/riscv.h, except that one hasn't been defined for RVE.  So probably should define one first.

sim/riscv/sim-main.c
```diff
@@ -1653,23 +2124,20 @@ void step_once (SIM_CPU *cpu)
...
-      sim_io_printf (sd, "sim: bad insn len %#x @ %#"PRIxTA": %#"PRIxTW"\n",
-		     len, pc, iw);
-      sim_engine_halt (sd, cpu, NULL, pc, sim_signalled, SIM_SIGILL);
-    }
 
-  iw |= ((unsigned_word)sim_core_read_aligned_2 (cpu, pc, exec_map, pc + 2) << 16);
+  if (len == 4)
+    iw |= ((unsigned_word)sim_core_read_aligned_2 (cpu, pc, exec_map, pc + 2) << 16);
+  else
+    iw |= ((unsigned_word)sim_core_read_aligned_2 (cpu, pc, exec_map, pc));
```
jim-wilson: This looks redundant.  For a 2-byte instruction, we already read two bytes.  No need to re-read the same 2 bytes again.

sim/riscv/sim-main.c
```diff
@@ -1678,11 +2146,26 @@ void step_once (SIM_CPU *cpu)
...
+      if ((op->pinfo & INSN_ALIAS))
+	continue;
+      /* Is this instruction restricted to a certain value of XLEN?  */
+      if (isdigit (op->subset[0]) && atoi (op->subset) != xlen)
+	continue;
+      /* It's a match.  */
+      pc = execute_one (cpu, iw, op);
+      break;
+
+      if ((op->match_func) (op, iw) && !(op->pinfo & INSN_ALIAS))
```
jim-wilson: Dead code after a break that should be removed.  This is the original code which is no longer useful.

sim/riscv/interp.c
```diff
@@ -154,13 +158,23 @@ sim_create_inferior (SIM_DESC sd, struct bfd *abfd,
...
+  Elf_Internal_Phdr *phdr;
+  int i, phnum;
 
   /* Set the PC.  */
   if (abfd != NULL)
     addr = bfd_get_start_address (abfd);
   else
     addr = 0;
   sim_pc_set (cpu, addr);
+  phdr = elf_tdata (abfd)->phdr;
```
jim-wilson: This assumes the highest address in the executable is the start of heap.  But if you use a linker script, you could have put special sections in high memory, and putting heap after them would be wrong.  Or alternatively, a board could have multiple memory regiions, and the region where the text/data is stored is not necessarily the highest address one.  Normally, one looks for the symbol _end and uses that as the start of heap.  But there is no one right way to do this.

sim/riscv/sim-main.c
```diff
@@ -817,6 +818,21 @@ execute_i (SIM_CPU *cpu, unsigned_word iw, const struct riscv_opcode *op)
   unsigned_word shamt_imm = ((iw >> OP_SH_SHAMT) & OP_MASK_SHAMT);
   unsigned_word tmp;
   sim_cia pc = cpu->pc + 4;
+  host_callback *cb = STATE_CALLBACK (sd);
```
jim-wilson: This block of code should go inside the MATCH_ECALL case where is the only place it is used.  It is wasted work if we don't have a syscall.  You can add a brace level for the case MATCH_ECALL for the new local variables.

sim/riscv/sim-main.c
jim-wilson: You put spaces around the first PRIiTW but not the second one. This is a minor nit.

sim/riscv/sim-main.c
jim-wilson: I would expect commas between the instruction operands. This is a minor nit. This occurs 4 times.

sim/riscv/sim-main.c
jim-wilson: This block of code should go inside the MATCH_ECALL case where is the only place it is used. It is wasted work if we don't have a syscall. You can add a brace level for the case MATCH_ECALL for the new local variables.

sim/riscv/interp.c
jim-wilson: This assumes the highest address in the executable is the start of heap. But if you use a linker script, you could have put special sections in high memory, and putting heap after them would be wrong. Or alternatively, a board could have multiple memory regiions, and the region where the text/data is stored is not necessarily the highest address one. Normally, one looks for the symbol _end and uses that as the start of heap. But there is no one right way to do this.

sim/riscv/sim-main.c
jim-wilson: This looks redundant. For a 2-byte instruction, we already read two bytes. No need to re-read the same 2 bytes again.

sim/riscv/sim-main.c
jim-wilson: Dead code after a break that should be removed. This is the original code which is no longer useful.

sim/riscv/sim-main.c
jim-wilson: Should use an EF_* value from include/elf/riscv.h, except that one hasn't been defined for RVE. So probably should define one first.

kito-cheng: Yeah, I think the problem is RVE support not merge to riscv-binutils-2.29 yet, I'll remove RVE in this PR and send another for this.

sim/riscv/interp.c
kito-cheng: I've update this part, it will use `_end` if possible.

sim/riscv/sim-main.c
kito-cheng: Fixed.

sim/riscv/sim-main.c
kito-cheng: Fixed.

sim/riscv/sim-main.c
kito-cheng: Fixed.

sim/riscv/sim-main.c
kito-cheng: Fixed.

sim/riscv/sim-main.c
kito-cheng: Fixed.

jim-wilson:
I am unfamiliar with github workflow in general, and riscv github workflow in particular.  I was assuming that the person who submitted the patch would merge them after approval.  This is how FSF binutils/gcc work.  If you are waiting for me to merge them, let me know.

jim-wilson:
I see that this is a request to add patches to riscv-binutils-2.29.  Should this be a request to add patches to riscv-next instead?  And then maybe cherry pick onto the riscv-binutils-2.29 branch?  There needs to be some clearly defined process here, or otherwise we end up with two different simulator ports for no reason.

palmer-dabbelt:
The problem with PRs against riscv-next is that github breaks them as riscv-next gets rebased.  I've been manually handling the impedance mismatch, but I think we might want to change our workflow a bit and do away with the "riscv-next" branch entirely.  I'm at the workshop, but let's talk when I'm back.

jim-wilson:
I was hoping for info from Palmer on how to manage riscv branches, but he has been busy, so I went ahead and merged this, and will deal with the fallout later.

PR #118, opened 2017-11-20
Rename sbadaddr/mbadaddr CSRs to stval/mtval

aswaterman:
The names changed in priv-1.10.  The old names will still be accepted as aliases.


palmer-dabbelt:
I pulled this into riscv-next, it'll eventually make it to 2.29.

PR #119, opened 2017-11-22
lw (a0) doesn't compress

palmer-dabbelt:
Don't merge this, as it's just a test case.  This is a low-priority bug, I'm just opening the PR so we don't forget about it.

### DIFF
--- a/gas/config/tc-riscv.c
+++ b/gas/config/tc-riscv.c
@@ -1185,6 +1185,25 @@ my_getSmallExpression (expressionS *ep, bfd_reloc_code_real_type *reloc,
   return reloc_index;
 }
 
+/* Detect and handle implicitly zero load-store offsets.  For example,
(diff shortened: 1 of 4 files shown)


aswaterman:
We should generalize the test to make sure the following all compress (since they use different patterns):

sw a0, (a0)
lw a0, (sp)
sw a0, (sp)

aswaterman:
I pushed a commit to this PR that fixes the bug.

jim-wilson:
I expanded the tests to include every case, and checked the patches into the FSF binutils tree.

jim-wilson:
Do I need to backport this patch, and if so, how exactly do I do that?

palmer-dabbelt:
I already did.  You can backport patches via

````
git checkout riscv-binutils-2.29
git cherry-pick HASH
````

you'll probably want git-merge-changelog, as otherwise you'll get a lot of conflicts.  I'm closing the PR, as it's fixed.

PR #122, opened 2017-12-12
Don't show registers that don't exist

timsifive:
The real fix here is to use the gdbarch_ functions instead of the riscv_ ones when displaying registers. Without that, the target description isn't checked which means that (sometimes) you'd get registers showing up that aren't actually on the target.

While discovering that this was the problem, I made some changes on the way that I think are beneficial to keep:
1. Add a few comments to undocumented gdb internal structures.
2. Factor out the register initialization into its own function.

### DIFF
--- a/gdb/riscv-tdep.c
+++ b/gdb/riscv-tdep.c
@@ -90,6 +90,8 @@ struct riscv_reg_info
   bool save_restore;
   const char *feature_name;
   struct reggroup *group;
+  // This register actually exists on the target.
+  bool exists;
 };
 
 static std::vector<struct riscv_reg_info> riscv_reg_info = {
@@ -163,6 +165,7 @@ static std::vector<struct riscv_reg_info> riscv_reg_info = {
 
   {RISCV_PRIV_REGNUM, {"priv"}, false, false, "org.gnu.gdb.riscv.virtual", general_reggroup},
 };
+// Map from fixed register number to register's info.
 static std::map<int, struct riscv_reg_info *> riscv_reg_map;
 
 struct register_alias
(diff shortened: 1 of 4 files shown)


ilg-ul:
> gdbarch_ functions instead of the riscv_ ones

does this affect gdb alone or gdb + gdb server?

timsifive:
I don't know what you mean by 'gdb server'.

ilg-ul:
openocd, j-link gdb server, qemu, ...

in other words, does this patch require the gdb servers to do something special, like sending the xml with the list of registers?

timsifive:
No. This change makes gdb work better when talking to servers that send an xml list. It does not affect servers that don't send a list.

ilg-ul:
that's good news, it means we can focus now on openocd, j-link gdb server, qemu, etc and convince them to send the right xml, without FP regs if not present, and especially with a realistic list of CSRs.

gdb/riscv-tdep.c
```diff
@@ -1362,19 +1266,155 @@ riscv_gdbarch_init (struct gdbarch_info info,
...
+
+  struct gdbarch_tdep *tdep;
+  const struct bfd_arch_info *binfo = info.bfd_arch_info;
+
+  int abi;
+
+  /* For now, base the abi on the elf class.  */
+  /* Allow the ELF class to override the register size. Ideally the target
+   * (OpenOCD/spike/...) would communicate the register size to gdb instead. */
+  abi = RISCV_ABI_FLAG_RV32I;
```
jim-wilson: This line appears unnecessary, as either abi is set below, or we call internal_error.

timsifive: Removed. Thanks for reviewing!

Issue #124, opened 2017-12-24
Linux build succeeds but Windows (cross) tools build fails


TM1234:
Overnight I tried to build the latest stuff from the repo and while the linux build completed the windows one failed:

```
x86_64-w64-mingw32-gcc -DHAVE_CONFIG_H -DWITH_DEFAULT_MODEL='"RV64G"' -DWITH_ALIGNMENT=NONSTRICT_ALIGNMENT -DWITH_TARGET_WORD_BITSIZE=64 -DWITH_TARGET_BYTE_ORDER=BFD_ENDIAN_LITTLE   -DDEFAULT_INLINE=0   -w      -I. -I/Host/Work/riscv64-unknown-elf/riscv-binutils-gdb.git/sim/riscv -I../common -I/Host/Work/riscv64-unknown-elf/riscv-binutils-gdb.git/sim/riscv/../common -I../../include -I/Host/Work/riscv64-unknown-elf/riscv-binutils-gdb.git/sim/riscv/../../include -I../../bfd -I/Host/Work/riscv64-unknown-elf/riscv-binutils-gdb.git/sim/riscv/../../bfd -I../../opcodes -I/Host/Work/riscv64-unknown-elf/riscv-binutils-gdb.git/sim/riscv/../../opcodes  -Wno-unknown-warning-option -Wno-extended-offsetof -Wno-deprecated-declarations -Wno-incompatible-pointer-types-discards-qualifiers -Wno-implicit-function-declaration -Wno-parentheses -Wno-format-nonliteral -Wno-shift-count-overflow -Wno-constant-logical-operand -Wno-shift-negative-value -Wno-format -pipe -ffunction-sections -fdata-sections -D__USE_MINGW_ACCESS -static-libstdc++ -static-libgcc -L/tmp/install/win64/lib -Wl,--gc-sections -Wl,--stack,12582912 -o run.exe \
  nrun.o libsim.a ../../bfd/libbfd.a ../../opcodes/libopcodes.a  ../../libiberty/libiberty.a -L../../zlib -lz  
libsim.a(sim-main.o):sim-main.c:(.text$execute_i+0x365b9): undefined reference to `link'
collect2: error: ld returned 1 exit status
make[3]: *** [run.exe] Error 1
Makefile:272: recipe for target 'run.exe' failed
make[3]: Leaving directory '/tmp/build/win64/binutils-gdb/sim/riscv'
make[2]: *** [all] Error 1
[...]

PR #125, opened 2017-12-28
Fix mingw builds on Windows.

jim-wilson:
Link is a POSIX system call that mingw does not support.  This causes build failures on Windows.  If we need this link support then it has to be done differently.  For now, we can just ifdef it out.


jim-wilson:
This fixes issue #124.

ilg-ul:
Although I'm not a reviewer, I think this fix should be applied before the next official release.

I just tried a build for the latest sources and the build failed on mingw-w64, as already reported in #124. As an workaround, I'll manually patch my sources, but the problem needs to be fixed.

ilg-ul:
Thank you, Jim!

PR #127, opened 2018-01-02
Default --with-expat=yes

timsifive:
XML support is required for OpenOCD to communicate the memory layout,
which is needed for gdb to treat flash like it should be.

Without this expat support, flashing will fail without any indication
why.

### DIFF
--- a/gdb/configure.ac
+++ b/gdb/configure.ac
@@ -667,7 +667,7 @@ GDB_AC_WITH_DIR([JIT_READER_DIR], [jit-reader-dir],
 
 AC_ARG_WITH(expat,
   AS_HELP_STRING([--with-expat], [include expat support (auto/yes/no)]),
-  [], [with_expat=auto])
+  [], [with_expat=yes])
 AC_MSG_CHECKING([whether to use expat])
 AC_MSG_RESULT([$with_expat])
 


timsifive:
@ilg-ul, can you confirm this is what you wanted?

ilg-ul:
I can confirm but... I was thinking on something else.

I was thinking to explicitly add `--with-expat` to the makefile used to build gdb, possibly with a comment mentioning that it is required by openocd.

jim-wilson:
This patch is unlikely to be accepted upstream.  We instead need something like what Liviu suggested.  A patch to the riscv-gnu-toolchain Makefile.in would work.

timsifive:
@ilg-ul If you know exactly what's required, can you create the pull request then?

ilg-ul:
I'm not that familiar with your build procedure to create the pull request, since I don't use it.

but I guess you need to add `--with-expat` somewhere after the line

https://github.com/riscv/riscv-gnu-toolchain/blob/d66c792dfe701c63a8df2028598174fbeee4e8f1/Makefile.in#L328

timsifive:
I'm not that familiar with your build procedure to create the pull request,
> since I don't use it.
>
I’m not familiar with it either. I just run make.

https://github.com/riscv/riscv-gnu-toolchain/blob/
> d66c792dfe701c63a8df2028598174fbeee4e8f1/Makefile.in#L328
>
> I don’t understand this code enough to make a sane pull request.

Tim
​

jim-wilson:
I can write a patch.

timsifive:
Thank you, Jim!

jim-wilson:
See  https://github.com/riscv/riscv-gnu-toolchain/pull/314

vapier:
no change like this would ever be acceptable upstream and i don't think even makes sense in any fork.  if you want to enable expat, do it in the scripts that are actually building gdb.  like Jim's commit.

PR #131, opened 2018-01-08
libdecnumber: Fix compilation of decNumber.c and decContext.c with separate build-dir

devurandom:
decnumber.o was built inside `$(srcdir)` and hence could not be found when building libdecnumber.a.  The same problem affects decContext.o.

### DIFF
--- a/libdecnumber/Makefile.in
+++ b/libdecnumber/Makefile.in
@@ -121,10 +121,10 @@ $(srcdir)/config.in: @MAINT@ $(srcdir)/configure
 
 # Dependencies.
 
-decContext.$(objext): decContext.c decContext.h decNumberLocal.h \
-	decContextSymbols.h
-decNumber.$(objext):  decNumber.c decNumber.h decContext.h decNumberLocal.h \
-	decNumberSymbols.h
+decContext.$(objext): $(srcdir)/decContext.c $(srcdir)/decContext.h $(srcdir)/decNumberLocal.h \
+	$(srcdir)/decContextSymbols.h
+decNumber.$(objext):  $(srcdir)/decNumber.c $(srcdir)/decNumber.h $(srcdir)/decContext.h $(srcdir)/decNumberLocal.h \
+	$(srcdir)/decNumberSymbols.h
 decimal32.$(objext):  $(srcdir)/$(enable_decimal_float)/decimal32.c \
    $(srcdir)/$(enable_decimal_float)/decimal32.h \
    $(srcdir)/$(enable_decimal_float)/decimal32Symbols.h \


jim-wilson:
Should be fixed upstream first, and then the fix pulled into our tree.  Upstream requires an FSF copyright assignment before patches can be accepted from a contributor.  So I'd prefer a bug report rather than a patch, unless you are willing to sign an FSF copyright assignment.

Upstream also won't accept a patch without an good explanation of why it is needed, and I don't see one here.  The Makefile uses VPATH, and hence there should be no need to add $(srcdir) for those two files.  Most people build gcc with a separate build dir, and no one has run into this problem other than you.  It isn't clear why you ran into this problem.  Some additional info describing what you are doing would help.  You should check the VPATH line in your libdecnumber Makefile to make sure it is OK.

It is wrong for decNumber.o to end up in the srcdir.  That doesn't happen when I build the gcc with a separate build dir.  This suggests that something went wrong on your end.  Maybe you did a build in the gcc source tree, and then tried to do a build with a separate build-dir?  That won't work without careful cleaning of the gcc source tree to remove all build artifacts.  Or maybe you are building on a target that is not well tested?  Having full host/target info and the configure command might help understand what is going wrong here.

devurandom:
The symptom that I noticed was that `decNumber.o` was not being build, but `ar` tried to include it in an archive. When I tried to build it manually, `make` told me that it was already up to date. I finally found it existing in the source directory.

I was working on https://github.com/dvc94ch/riscv-rust-toolchain, getting it to work with newer LLVM and Rust, when I ran into this. The target is `riscv32-unknown-elf` and the configure command was:
```
cd $(CURDIR)/build && $(CURDIR)/../riscv-binutils-gdb/configure \
                --target=riscv32-unknown-elf \
                --prefix=$(CURDIR)/../toolchain \
                --disable-werror \
                --disable-debug \
                --without-guile \
                --enable-python \
                CC="/usr/bin/ccache cc") \
                CXX="/usr/bin/ccache cxx")
```

But it changed a little bit (addition of `CC`/`CXX`/ccache, `--without-guild`, `--enable-python`) during my journey, until I realised that the source dir was messed up in this way, so I cannot ATM be sure what caused the build to go wrong.

`VPATH` in the `Makefile` (as generated from `Makefile.in` inside the build dir) appears to be correct.

Why would the other `# Dependencies` need an explicit `$(srcdir)` in their dependencies, but not `decNumber.c` and `decContext.c`?

I would not mind assigning copyright of my work on this to the FSF, as far as this is legally possible for me, but [the process](https://www.fsf.org/licensing/assigning.html) seems to be rather involved, so it might take a while.

jim-wilson:
The other files aren't in srcdir; they are in a subdir of srcdir with no Makefile.  Hence they need explicit paths.  I would also point to Makefiles in other dirs that use files without srcdir, such as the reference to errors.c in gcc/Makefile.in.  I would also point out that many thousands of people have built the software without running into this problem.

I think it is more likely that you made a mistake somewhere along the way.  I can reproduce your error if I configure in the riscv-binutils-gdb dir, build all-libdecnumber, remove libdecnumber/libdecnumber.a, configure in a separate build dir, and then build all-decnumber again.  I get an ar error for a missing decNumber.o file.  This is user error.  You must remove all build artifacts from the source dir after a config in the source dir before a config with a separate build dir will work.

The gcc makefiles actually have some protections to avoid this problem.  Host libraries are built in a subdir host-$host instead of in the library source dir, and configure emits an error if you have build artifacts in the source dir.  Binutils unfortunately does not have these checks.

I can't think of another way that you can run into this problem, as otherwise nothing should every be placed in the source tree when you config with a separate build dir.

devurandom:
Thanks for investing the time to explain this!

PR #132, opened 2018-01-08
Ignore errors when reading register to save them

timsifive:
Debug servers may not be able to determine exactly which registers exist
on a target, meaning in turn that gdb can't be sure. When gdb goes to
save these registers before a function call, failure to read one of them
should not be fatal.

This change ignores these failures. Registers that were not read, will
not be restored.

Unfortunately for each register that can't be read there will be a
message to the user. Getting rid of that message only for this use case
is not trivial, so I'm punting on it for now.

### DIFF
--- a/gdb/regcache.c
+++ b/gdb/regcache.c
@@ -323,6 +323,21 @@ regcache_save (struct regcache *regcache,
   regcache->save (cooked_read, src);
 }
 
+struct catch_cooked_read_args {
+    enum register_status status;  // result of function call
+    regcache_cooked_read_ftype *cooked_read;
+    void *src;
+    int regnum;
+    gdb_byte *buf;
+};
+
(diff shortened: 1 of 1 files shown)


ilg-ul:
How did you verify that this new change in the common OpenOCD code does not affect other targets? Can this patch be upstreamed?

timsifive:
I did not verify that this change to *gdb* does not affect other targets. What effects are you concerned about?

I don't know if upstream will accept this or not. Hopefully somebody with more experience can make a better guess, but otherwise the only option is to try and see.

ilg-ul:
yes, to gdb, sorry (too many late hours in the last weeks...).

I'm always concerned when I see changes performed to non risc-v specific files, both to gdb and especially to openocd.

PR #140, opened 2018-02-27
Added hcos OS to all building riscv64-unknown-hcos target.

rishikhan:


jim-wilson:
Adding patches to binutils-2.29 isn't useful.  I think it would be better to submit patches upstream.

rishikhan:
ok. Just https://sourceware.org/git/gitweb.cgi?p=binutils-gdb.git ? How will that come back down to riscv-gnu-toolchain?

vapier:
you want to send all config.{sub,guess} patches to https://savannah.gnu.org/projects/config

jim-wilson:
On Tue, Feb 27, 2018 at 12:57 PM, Rishi Khan <notifications@github.com>
wrote:

> ok. Just https://sourceware.org/git/gitweb.cgi?p=binutils-gdb.git ? How
> will that come back down to riscv-gnu-toolchain?
>

Email patches to binutils@sourceware.org.

You can see a mailing list archive at
     https://sourceware.org/ml/binutils/
Here is an IA-64 patch I sent yesterday.
    https://sourceware.org/ml/binutils/2018-02/msg00438.html
Here is a RISC-V patch from the MIPS maintainer.
    https://sourceware.org/ml/binutils/2018-02/msg00029.html

I normally use "git format-patch" to create the patches, and "git
send-email" to mail them to the list.  You can edit the patch file after
git format-patch creates it, and before you mail it, to add some info about
the patch, e.g. what it fixes, how you tested it, etc.

Jim

jim-wilson:
I should have looked at the patch first.  Only binutils patches can be sent to the binutils mailing list.  Some things in the binutils/gcc trees are imported from elsewhere.  The GCC web site has a list at
    https://gcc.gnu.org/codingconventions.html#upstream
and it mentions that config.sub patches need to go to config-patches@gnu.org.

There is a lot of general info about how to make changes to gcc can be found at
    https://gcc.gnu.org/svnwrite.html
The rules for binutils are similar, but much simpler, as it is a smaller project with fewer contributors.  The binutils rules are just not documented as well as the gcc rules, so the gcc rules are probably a good reference to start with.

rishikhan:
This was successfully submitted to the config.sub repo. Can we merge it here now I I can use this as my base?

See:
http://git.savannah.gnu.org/gitweb/?p=config.git;a=commitdiff;h=b69837cbbd86bca2210768deb5e9837f366d166b

vapier:
if it's in the GNU config project, then syncing to upstream gcc/binutils is usually pretty easy.  send a patch or just e-mail them and ask them to do it.  or just wait and eventually someone will do it.

rishikhan:
I'm sure it will get synced soon ... but who will sync THIS repo?

jim-wilson:
It's complicated.

We'd like to use official FSF sources instead of the ad-hoc branches we have now.  Binutils, gcc, and newlib are upstream, so that should be pretty easy, but changes need to be verified and tested and that takes time.  Gdb and qemu are not yet upstream, but expected to appear upstream shortly, so waiting for that may be beneficial. 

Once gdb is upstream we have a new problem, which is that binutils and gdb share a git tree, but have separate release branches and release schedules, with release branches that contain different versions of common files, and that are missing files used only by the other package.  This means we will need two sources trees, one for binutils and one for gdb, which requires build script changes, and more testing and verification.

GCC is currently gcc-7.3, but a gcc-8 release will happen in a few months, and then we have the problem that some people will want to remain with gcc-7.3 for stability, and some people will want to move to gcc-8 for new features, and we can't easily satisfy both with a single tree.  We could try adding configure support to choose the gcc version, but supporting two gcc releases is twice the work, and this could be a problem that may result in a less stable tree.  Eventually we will also have version issues with other tools too.

There are also people who will want certain patches backported from the mainline development tree into release branches, and other people who may not want certain patches backported.  We may run into difficult to resolve conflicts.

There are also issues with other packages that people may want to add, other languages that people may want to support, questions about which configure options should be on by default, etc.

Long term, the idea that riscv-gnu-toolchain is the definitive toolchain that everyone should be using is just not going to work.  Anyone serious about building toolchains should probably be looking at maintaining their own build scripts and patch sets to build the toolchain the way they need it.

I think riscv-gnu-toolchain does have value as an easy way to get started with RISC-V development, for people who don't care about the exact versions of tools, the exact set of tools, the exact set of patches, the exact configure options, etc.  And if you do care about that stuff, then you should probably move on to building your own toolchains the way you need them.

vapier:
Jim nailed it!

PR #143, opened 2018-03-27
Complete full reg list if some aren't readable

timsifive:
Previously `info all-registers` and `info registers all` would terminate
as soon as it got to a register that was not readable. Now the user gets
a message and the register list continues.

### DIFF
--- a/gdb/riscv-tdep.c
+++ b/gdb/riscv-tdep.c
@@ -538,7 +538,20 @@ riscv_print_register_formatted (struct ui_file *file, struct frame_info *frame,
       unsigned long long d;
       int prefer_alias = regnum >= RISCV_FIRST_CSR_REGNUM;
 
-      if (!deprecated_frame_register_read (frame, regnum, raw_buffer))
+      int read_result = -1;
+      TRY
+        {
+          read_result = deprecated_frame_register_read (frame, regnum, raw_buffer);
+        }
+      CATCH (ex, RETURN_MASK_ERROR)
+        {
+          fprintf_filtered (file, "%-15s%s\n",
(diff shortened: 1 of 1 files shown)


jim-wilson:
The upstream code is different.  It no longer uses deprecated_frame_register_read.  It might be useful to file a bug report upstream to get that copy fixed also, assuming it has the same problem.

I don't see similar changes in other ports.  However, I'm guessing the problem here is that we have a long list of CSRs that most parts only implement some of, and the only way to tell if a CSR exists is to try reading from it, which means we need to be able to ignore reads from registers that don't exist.  That would be a problem other ports don't have.  If this is the problem you are working around, then the patch looks reasonable.

timsifive:
Should I be making PRs against an upstream branch? Can you point me to
which one?

Mainline OpenOCD does not propagate errors that occur when reading a
register, which might be one reason that other targets don't do this. It
probably also doesn't matter for them since they don't have the problem of
having registers that it's not possible to detect whether they exist or not
(without possible side effects). But I want RISC-V OpenOCD to report errors
when a non-existent CSR is requested, and this makes gdb deal with that
cleanly.

jim-wilson:
Upstream gdb is at
    https://sourceware.org/gdb/
It only has the base RISC-V port at the moment, so it probably isn't very useful, but eventually this should be a full port, and will replace the github riscv-binutils-gdb port.  The port is on mainline, added after the gdb-8.1 release, so not in any FSF GDB release as yet.  I don't know where the development work is being done.  I suspect the developer, Andrew Burgess, has his own private tree somewhere.  But the riscv support is there in gdb mainline, so we can submit bug reports against it.

PR #146, opened 2018-04-25
Assemble and disassemble proposed FENCE.TSO instruction

aswaterman:
I'm putting this here as a placeholder rather than submitting the patch to FSF because it's still just a proposal at this point.

FENCE.TSO is analogous to Power's ```lwsync``` and is a performant mapping for ```smp_store_release``` in the Linux kernel and ```atomic_thread_fence(memory_order_acq_rel)``` in C.  It executes correctly on existing implementations because it is encoded in the subspace of FENCE RW,RW, which is strictly stronger.

cc @palmer-dabbelt @jim-wilson

### DIFF
--- a/include/opcode/riscv-opc.h
+++ b/include/opcode/riscv-opc.h
@@ -141,6 +141,8 @@
 #define MASK_FENCE  0x707f
 #define MATCH_FENCE_I 0x100f
 #define MASK_FENCE_I  0x707f
+#define MATCH_FENCE_TSO 0x8330000f
+#define MASK_FENCE_TSO  0xfff0707f
 #define MATCH_MUL 0x2000033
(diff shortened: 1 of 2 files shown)


palmer-dabbelt:
LGTM.  IIRC binutils isn't releasing for a while so there's no rush here, so assuming this is going to land in an actual released spec soon I think we should wait for that.  It's a bit pedantic, but I think it's a good precedent to set as we've been a bit sloppy about that in the past.

aswaterman:
Agreed.

aswaterman:
The memory consistency model has been ratified, so it's probably time to send this upstream.

palmer-dabbelt:
https://sourceware.org/ml/binutils/2018-10/msg00012.html

PR #155, opened 2018-08-11
RISC-V: New emulations to make path searches follow glibc ABI.

jim-wilson:
This cherry picks a patch from the FSF Binutils tree which is required by an FSF GCC patch that was already backported earlier to riscv-gcc.  This fixes a build problem with linux multilib toolchains, reported in riscv-gnu-toolchain issue 351.  Tested with a regression run in riscv-gnu-toolchain.


palmer-dabbelt:
Thanks!

PR #159, opened 2018-09-07
Fix output for 'info registers' command.

vitaly-gaiduk:
Until the fix you could see output of value priv register like below:
```
(gdb) info registers priv
priv           0x01754903       prv:3 [Machine]
```
This is because the size of output is set to four bytes, but priv register contains pirvilege mode in last byte and if you want to parse hex value (e.g. in some your test case in riscv-tests/debug/gdbserver.py) you should to add workaround for getting right hex value (e.g. masking this low-byte or smth).

After fix you will be able to parsing like other registers:
```
(gdb) info registers priv
priv           0x03     prv:3 [Machine]
```

### DIFF
--- a/gdb/riscv-tdep.c
+++ b/gdb/riscv-tdep.c
@@ -578,9 +578,28 @@ riscv_print_register_formatted (struct ui_file *file, struct frame_info *frame,
 
       size = register_size (gdbarch, regnum);
       get_formatted_print_options (&opts, 'x');
+
+      char csize;
+      switch (size) {
+          case 1:
+            csize = 'b';
+            break;
+          case 2:
+            csize = 'h';
+            break;
+          case 4:
+            csize = 'w';
+            break;
+          case 8:
+            csize = 'g';
+            break;
+          default:
+            csize = 0;
+      }
+
       print_scalar_formatted (raw_buffer + offset,
 			      register_type (gdbarch, regnum), &opts,
-			      size == 8 ? 'g' : 'w', file);
+			      csize, file);
       fprintf_filtered (file, "\t");
       if (size == 4 && riscv_isa_regsize (gdbarch) == 8)
 	fprintf_filtered (file, "\t");
--- a/gdb/riscv-tdep.c
+++ b/gdb/riscv-tdep.c
@@ -594,7 +594,7 @@ riscv_print_register_formatted (struct ui_file *file, struct frame_info *frame,
             csize = 'g';
             break;
           default:
-            csize = 0;
+            internal_error (__FILE__, __LINE__, _("unknown size for register"));
       }
 
       print_scalar_formatted (raw_buffer + offset,


gdb/riscv-tdep.c
```diff
@@ -578,9 +578,28 @@ riscv_print_register_formatted (struct ui_file *file, struct frame_info *frame,
...
+            csize = 'h';
+            break;
+          case 4:
+            csize = 'w';
+            break;
+          case 8:
+            csize = 'g';
+            break;
+          default:
+            csize = 0;
```
timsifive: Can you error() here? Otherwise you'll just end up with an error inside print_scalar_formatted, which will be harder to debug.

jim-wilson:
This would be more useful in the FSF gdb sources.  Patches added to riscv-gnu-toolchain may be lost when this gdb is replaced with the upstream gdb.  FSF gdb patches require a copyright assignment.  If you don't have one, it might be  simpler to just report a bug in the FSF gdb bugzilla, and let someone else with a copyright assignment write a patch.

vitaly-gaiduk:
@jim-wilson I've looked at gdb branches (master and binutils-2_29_1) but could not find such version of riscv-tdep.c. Could you write me the branch which you merge from upstream?

jim-wilson:
We have two gdb ports.  One in the FSF tree, and one in the riscv-binutils-gdb tree.  There is no merging as yet.  At some point, the riscv-binutils-gdb one will be thrown away and we will use the FSF ome instead.  See https://sourceware.org/gdb/

jim-wilson:
Note that we can't submit patches to the FSF that we don't have copyright assignments for.  You really should not be accepting patches from anyone without verifying copyright assignment status first.  This means that we can no longer trust any patches in the riscv-gnu-toolchain gdb tree, and can no longer submit any upstream.  This is very bad practice and is going to cause much trouble.  I am considering reverting the patch to avoid problems.

timsifive:
@vitaly-gaiduk, Jim is right in that we need a copyright assignment to avoid any problems with contributing this code upstream. https://sourceware.org/git/gitweb.cgi?p=binutils-gdb.git;a=blob;f=gdb/CONTRIBUTE;hb=HEAD#l43 outlines the policy requirements, but I can't find a simple document that explains what is required for copyright assignment. @jim-wilson, do you have a link to something that details the process. (I assume it's something like, print out, sign, scan, e-mail.)

jim-wilson:
The preferred approach is to send email to assign@gnu.org, mention what project you are contributing to, what kinds of contributions you are making, etc, and wait for a response.  The FSF may ask a few questions, and then send you the appropriate form.  If you have an employer, then you will either need a copyright assignment from your employer, or else a disclaimer from your employer and a personal assignment.  Either approach requires a VP level signature.  Depending on who you work for, this may be easy to impossible.  If you are student, you may need paperwork from the university you are attending.  Etc.  The FSF relies on volunteer labor, so the assignment process sometimes takes a while.  if you are lucky, then it should be done in a month or two.  If you are unlucky, e.g. if your employer doesn't cooperate, then it can take over a year, and may never be completed.

You can always file a bug report, and let someone else who has an assignment fix it.  The preferred place for such a bug report is the FSF GDB Bugzilla.
https://sourceware.org/gdb/bugs/

vitaly-gaiduk:
Thanks a lot @jim-wilson and @timsifive for detailed explanation!

PR #160, opened 2018-09-15
RISC-V: enable have_nonsteppable_watchpoint by default

craigblackmore:
This is a fix for https://github.com/riscv/riscv-openocd/issues/295

The RISC-V debug spec 0.13 recommends that write triggers fire before
the write is committed. If the target follows this behaviour, then
have_nonsteppable_watchpoint needs to be set to 1 so that GDB will step
over the watchpoint before checking if the value has changed.

This patch adds a setshow for have_nonsteppable_watchpoint which defaults
to 1 to match the recommended behaviour. If a target does not follow
this timing, then 'set riscv have_nonsteppable_watchpoint 0' will need
to be issued on the command line.

Note: I will also submit a patch to upstream GDB

### DIFF
--- a/gdb/riscv-tdep.c
+++ b/gdb/riscv-tdep.c
@@ -172,6 +172,20 @@ struct register_alias
   int regnum;
 };
 
+static int riscv_have_nonsteppable_watchpoint = 1;
+
+/* The set callback for 'set riscv have-nonsteppable-watchpoint'. */
+
+static void
+set_have_nonsteppable_watchpoint (char *args, int from_tty,
+			       struct cmd_list_element *c)
+{
+  struct gdbarch *gdbarch = target_gdbarch ();
+
+  set_gdbarch_have_nonsteppable_watchpoint(gdbarch,
+					   riscv_have_nonsteppable_watchpoint);
+}
+
 static enum auto_boolean use_compressed_breakpoints;
 /*
 static void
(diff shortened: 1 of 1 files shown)


craigblackmore:
```make all-tests``` passes, but ```make all``` fails because pylint complains about some of the testing modules in ```riscv-tests/debug```:

```
$ make all
pylint --rcfile=pylint.rc `git ls-files '*.py'`
Using config file /home/craig/work/projects/sifive/riscv-tools/riscv-tests/debug/pylint.rc
************* Module gdbserver
R:396, 4: Either all return statements in a function should return an expression, or none of them should. (inconsistent-return-statements)
R:416, 4: Either all return statements in a function should return an expression, or none of them should. (inconsistent-return-statements)
R:653, 4: Either all return statements in a function should return an expression, or none of them should. (inconsistent-return-statements)
R:681, 4: Either all return statements in a function should return an expression, or none of them should. (inconsistent-return-statements)
R:1020, 4: Either all return statements in a function should return an expression, or none of them should. (inconsistent-return-statements)
************* Module targets
R: 44, 8: Unnecessary "else" after "return" (no-else-return)
************* Module testlib
W:466,27: Name async will become a keyword in Python 3.7 (assign-to-new-keyword)
W:474,12: Name async will become a keyword in Python 3.7 (assign-to-new-keyword)
W:476,12: Name async will become a keyword in Python 3.7 (assign-to-new-keyword)
R:466, 4: Either all return statements in a function should return an expression, or none of them should. (inconsistent-return-statements)
R:531, 8: Unnecessary "else" after "return" (no-else-return)

------------------------------------------------------------------
Your code has been rated at 9.93/10 (previous run: 9.93/10, +0.00)

Makefile:14: recipe for target 'pylint' failed
make: *** [pylint] Error 12
```

timsifive:
Thanks for checking riscv-tests/debug. Unfortunately different versions of pylint perform different checks. Hopefully I'll find some time to address the ones your version found later.

PR #162, opened 2018-10-17
Back-port from trunk to 2.31.1

kito-cheng:
Back-port list:
 - RISC-V: Adjust __global_pointer$ value to reduce code size.
 - RISC-V: Delete zero-size .tdata.dyn section.
 - RISC-V: Add fence.tso instruction
 - RISC-V: Pc-rel to gp-rel relaxation function cleanup.
 - RISC-V: Give error for RVE PLTs.
 - RISC-V: For PCREL_LO12, fix addend handling in auipc lookup.
 - RISC-V: Allow pcrel_lo addends, error on addend overflow.
 - RISC-V: Fix library search path for rv32.
 - RISC-V: bge[u] should get higher priority than ble[u].
 - RISC-V: Correct the requirement of compressed floating point instructions
 - RISC-V: Allow instruction require more than one extension
 - RISC-V: Reject empty rouding mode and fence operand.
 - RISC-V: Fix TLS and --gc-sections conflict.
 - RISC-V: Set insn info fields correctly when disassembling.
 - RISC-V: Fix gas configure support for riscv*-*-*.
Tested with RV32GC and RV64GC.


palmer-dabbelt:
Thanks, Kito!

PR #163, opened 2018-12-10
fix a typo in the assembler

benshi001:
fix a typo in the error message of the assembler.


jim-wilson:
This isn't a typo, it is a deliberate style choice in binutils to use a back-quote paired with a single quote in error messages.

PR #178, opened 2019-09-30
z subset parsing V2

mablinov:
I've taken the previous branch, rebased it on the now newly rebased riscv-bitmanip master branch, and squashed the small bugfix commits. It ended up being a bit ugly because half of it was already upstream, but I think it should be ok.

Those tests that it does not pass are tests that test the old-style of parsing, so it is correct that they are now failing. These are:

gas/riscv/march-ok-s-with-version
gas/riscv/march-ok-s
gas/riscv/march-ok-sx

I will ofcourse fix these, its just that I instinctively pressed control-enter, which opened the pull request as I was midway writing it...


bfd/elfnn-riscv.c
```diff
@@ -2960,14 +2979,9 @@ riscv_merge_arch_attr_info (bfd *ibfd, char *in_arch, char *out_arch)
...
-  if (!riscv_merge_non_std_and_sv_ext (ibfd, &in, &out, riscv_non_std_ext_p))
-    return NULL;
-  /* Merge standard supervisor extension.  */
-  if (!riscv_merge_non_std_and_sv_ext (ibfd, &in, &out, riscv_std_sv_ext_p))
-    return NULL;
-  /* Merge non-standard supervisor extension.  */
-  if (!riscv_merge_non_std_and_sv_ext (ibfd, &in, &out, riscv_non_std_sv_ext_p))
+
+  /* Merge all non-single letter extensions with single call.  */
+  if (!riscv_merge_non_std_ext (ibfd, &in, &out))
```
kito-cheng: `riscv_merge_non_std_ext` rename to `riscv_merge_multi_letter_ext` might be better? Z* isn't non-standard things like your comment, it's non-single letter extensions.

bfd/elfxx-riscv.c
```diff
@@ -1294,21 +1324,154 @@ riscv_parse_sv_or_non_std_ext (riscv_parse_subset_t *rps,
...
+  return -1;
+}
+
+/* Predicator function for x-prefixed extensions.
+   Anything goes, except the literal 'x'.  */
+
+static bfd_boolean
+riscv_ext_x_valid_p (const char *arg)
+{
+  if (!strcasecmp (arg, "x"))
```
kito-cheng: Could you add `riscv_non_std_ext_strtab` like `riscv_std_s_ext_strtab`? so that we can simplify those parsing logic into common single function.
e.g.
```
static bfd_boolean
riscv_ext_z_valid_p (const char *arg)
{
   return riscv_multi_letter_ext_valid_p (arg, riscv_z_ext_strtab);
}
static bfd_boolean
riscv_ext_s_valid_p (const char *arg)
{
   return riscv_multi_letter_ext_valid_p (arg, riscv_s_ext_strtab);
}
static bfd_boolean
riscv_ext_x_valid_p (const char *arg)
{
   return riscv_multi_letter_ext_valid_p (arg, riscv_non_std_ext_strtab);
}
```

bfd/elfxx-riscv.c
```diff
@@ -1294,21 +1324,154 @@ riscv_parse_sv_or_non_std_ext (riscv_parse_subset_t *rps,
...
+const char * const riscv_std_s_ext_strtab[] =
+  {
+   NULL
+  };
+
+/* Find the index corresponding to the z-extension
+   name `ext'. If not found, return -1.  */
+
+int
+riscv_std_z_ext_index (const char *ext)
```
kito-cheng: Seems like nobody call `riscv_std_z_ext_index`?

bfd/elfxx-riscv.c
mablinov: I think we can't do a table lookup for `x`, because we don't know ahead of time what arbitrary extensions the user may wish to use. so `riscv_ext_x_valid_p` can only check that it is prefixed with `x`.

kito-cheng: Oh, my point was that make new `x` extension easier to add if user want, we can just put a stub here.
But I didn't notice we didn't check any `x` extension, and just bypass that, so that might cause different behavior for parsing `x` according my review comment.

So I am ok for your `riscv_ext_x_valid_p` now :)

bfd/elfxx-riscv.c
```diff
@@ -1294,21 +1324,154 @@ riscv_parse_sv_or_non_std_ext (riscv_parse_subset_t *rps,
...
+    return FALSE;
+
+  return TRUE;
+}
+
+/* Predicator functions for z-prefixed extensions.
+   Only known z-extensions are permitted.  */
+
+static bfd_boolean
+riscv_ext_z_valid_p (const char *arg)
```
kito-cheng: However I think riscv_ext_z_valid_p and riscv_ext_s_valid_p need to using some common utils function to check, otherwise we'll have 3 almost copy/paste code after supporting h* extension.

like
```
static bfd_boolean
riscv_ext_z_valid_p (const char *arg)
{
   return riscv_multi_letter_ext_valid_p (arg, riscv_z_ext_strtab);
}
static bfd_boolean
riscv_ext_s_valid_p (const char *arg)
{
   return riscv_multi_letter_ext_valid_p (arg, riscv_s_ext_strtab);
}
```

include/opcode/riscv.h
```diff
@@ -293,7 +293,7 @@ static const char * const riscv_pred_succ[16] =
 
 /* The maximal number of subset can be required. */
 #define MAX_SUBSET_NUM 4
-
+
```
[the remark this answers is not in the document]
mablinov: Hi Kito, sorry I haven't been keeping up too much with standard developments - `sx` is now gone, I take it?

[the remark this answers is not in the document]
kito-cheng: Yes, it's gone on latest spec, and checked it's OK to remove with LLVM and ISA guys.

mablinov:
Thanks for reviewing Kito, I've incorporated your changes.

mablinov:
> Do you mind split Z* paring support for -march and removing sx into standalone patch and upstream that?

You mean, strip out the bitmanip stuff, upstream that, merge back into here, and then apply bitmanip-specifics?

bfd/elfxx-riscv.c
```diff
@@ -1237,28 +1237,58 @@ riscv_parse_std_ext (riscv_parse_subset_t *rps,
...
+   riscv_isa_ext_class_t. One of these should be provided for each
+   possible class, except RV_ISA_CLASS_UNKNOWN.  */
+
+typedef struct riscv_parse_config
+{
+  /* Class of the extension. */
+  riscv_isa_ext_class_t class;
+
+  /* Lower-case prefix string for error printing
+     and internal parser usage, e.g. "sx", "z".  */
```
kito-cheng: `sx` is gone :P

bfd/elfxx-riscv.c
```diff
@@ -1237,28 +1237,58 @@ riscv_parse_std_ext (riscv_parse_subset_t *rps,
...
+  const char *prefix;
+
+  /* Predicate which is used for checking whether
+     this is a "known" extension. For 'x' and 'sx',
```
kito-cheng: Ditto.

bfd/elfxx-riscv.c
```diff
@@ -1294,21 +1321,122 @@ riscv_parse_sv_or_non_std_ext (riscv_parse_subset_t *rps,
 
       *q = '\0';
 
+      /* Check that the name is valid.
+	 For 'x', anything goes but it cannot simply be 'x'.
+	 For 'z', it must be known from a list and also cannot simply be 'z'.
+	 For 's', it must be known from a list and also *can* simply be 's'.
+	 For 'sx', see 'x'.  */
```
kito-cheng: Ditto.

kito-cheng:
>> Do you mind split Z* paring support for -march and removing sx into standalone patch and 
 upstream that?

> Do you mean: strip out the bitmanip stuff, upstream that, merge back into here, and then apply bitmanip-specifics?

Yes, because I think other extension might need that, like Ztso, Zam, Zifencei, Zicsr in latest unprivileged spec, and Vector extension also defined some Zv* extensions.

It would be great if this part go upstream first if you have time :)

cliffordwolf:
> I agree with that. I'd like to have as little non-bitmanip specific code introduced through this branch as possible.

But we can merge this now and I'll just rebase our branch once the non-bitmanip functionality was upstreamed independently. Just let me know what you'd prefer @mablinov.

mablinov:
@cliffordwolf I would really appreciate it if you guys could merge it into bitmanip. I'll send it upstream, but this is a major user-facing change that will probably take a lot of time to convince people to move over to.

PR #179, opened 2019-10-07
Support more vsew for vtypei setting.

Nelson1225:
Dear all,

I get a request that we need to support more vsew registers for vtypei setting of vsetvli.

According to riscv vector spec: 
https://github.com/riscv/riscv-v-spec/blob/master/v-spec.adoc#321-vector-standard-element-width-vsew

vsew[2:0] 3-bit, 8 types of vsew:
e8, e16, e32, e64, e128, e256, e512, e1024

Thanks
Best Regards
Nelson


gas/testsuite/gas/riscv/vector-insns.s
```diff
@@ -2,6 +2,7 @@
 	vsetvli a0, a1, 0
 	vsetvli a0, a1, 0x7ff
 	vsetvli a0, a1, e16,m2,d4
+	vsetvli a0, a1, e1024, m8, d8
```
kito-cheng: Could you add more test case to cover all new values?

Nelson1225: OK, I have added the new cases for the new values (e256, e512, e1024).

As for the test cases, which can cover all possible usage, we will support them in another commit :)

PR #184, opened 2019-11-08
Riscv binutils 2.33.1 rvv 0.8.x

Nelson1225:
There are two commits, the first one is used to support vector format for .insn directive, and the second one is used to support INSN_CLASS_F (F_AND_C and V_AND_F) for the instructions that use the FPR.  The second commit should be a upstream issue rather than the vector specific issue, so I will send it to upstream first, and then backport it on the vector branch.

Vector instruction are based on the A-extension, and the A-ext instruction is based on R-type instruction.  The OPIVI instruction is confusing, I suppose this type instruction is based on R-type, but the Rt field must be a signed immediate rather than a register.  The vector memory unit-stride instructions have the same issue (lumop and sumop) as OPIVI.  Currently, I use the register to represent the simm4 and [l|s]umop for OPIVI and unit-stride, that is, 

// OPIVI (rs (rs1) is simm4)
// x31 = 0x11111 = -1 (two's complement)
.insn r 0x57, 0x3, 0x0, v11, x0, a3
.insn r 0x57, 0x3, 0x0, v11, x31, a3
// unit-stride (rt (rs2) = lumop and sumop)
// x16 = 0x10000
.insn r LOAD_FP, 0x0, 0x0, v11, a2, x0
.insn r LOAD_FP, 0x0, 0x1, v11, a2, x16

// OPIVI
vadd.vi  v11,v13,0,v0.t
vadd.vi  v11,v13,-1,v0.t
// unit-stride
vlbu.v[       ]+v11,(a2),v0.t
vlbuff.v[     ]+v11,(a2)

Thanks
Best Regards
Nelson


jim-wilson:
There is an open issue where I ask about instruction format names.
https://github.com/riscv/riscv-v-spec/issues/93
This did come up in one of the vector extension task group meetings, but it isn't a high priority issue so I don't think that Krste has tried to fix it yet.

opcodes/riscv-opc.c
```diff
@@ -1575,96 +1575,116 @@ const struct riscv_opcode riscv_insn_types[] =
...
-{"r",       0, INSN_CLASS_I,  "O4,F3,F7,D,s,T",     0,    0,  match_opcode, 0 },
-{"r",       0, INSN_CLASS_I,  "O4,F3,F7,d,S,T",     0,    0,  match_opcode, 0 },
-{"r",       0, INSN_CLASS_I,  "O4,F3,F7,D,S,T",     0,    0,  match_opcode, 0 },
+{"r",       0, INSN_CLASS_F,  "O4,F3,F7,D,s,t",     0,    0,  match_opcode, 0 },
+{"r",       0, INSN_CLASS_F,  "O4,F3,F7,d,S,t",     0,    0,  match_opcode, 0 },
+{"r",       0, INSN_CLASS_F,  "O4,F3,F7,D,S,t",     0,    0,  match_opcode, 0 },
+{"r",       0, INSN_CLASS_F,  "O4,F3,F7,d,s,T",     0,    0,  match_opcode, 0 },
+{"r",       0, INSN_CLASS_F,  "O4,F3,F7,D,s,T",     0,    0,  match_opcode, 0 },
+{"r",       0, INSN_CLASS_F,  "O4,F3,F7,d,S,T",     0,    0,  match_opcode, 0 },
+{"r",       0, INSN_CLASS_F,  "O4,F3,F7,D,S,T",     0,    0,  match_opcode, 0 },
```
jim-wilson: This looks right, but isn't vector specific.  Trying to maintain this on the vector branch could lead to merging problems later if someone else submits a patch upstream to change the .insn support.  Maybe this stuff should just go upstream now?  And then you can backport it if you need it on the vector branch.  Likewise for the other non-vector specific changes.

Nelson1225: Yes your right, this should be a upstream issue rather than the vector specific issue.  Sure I can send the patch to upstream now, thanks for the suggestion :)

PR #185, opened 2019-11-10
ld/ldlang: Fix linker relaxation failure if there are R_RISCV_ALIGN type relocations on RISC-V

yiting16:
See bug report at https://sourceware.org/bugzilla/show_bug.cgi?id=25181

    ld/
        * ldlang.c (lang_size_sections_1): split the call to
          size_input_section() to 2 parts to fix linker relaxation
          failure if there are R_RISCV_ALIGN type relocations on RISC-V
        * testsuite/ld-riscv-elf/ld-riscv-elf.exp: add a test case
        * testsuite/ld-riscv-elf/relax*: the test case files

### DIFF
--- a/ld/ldlang.c
+++ b/ld/ldlang.c
@@ -4882,25 +4882,23 @@ insert_pad (lang_statement_union_type **ptr,
 /* Work out how much this section will move the dot point.  */
 
 static bfd_vma
-size_input_section
+size_input_section_part1
   (lang_statement_union_type **this_ptr,
    lang_output_section_statement_type *output_section_statement,
-   fill_type *fill,
-   bfd_vma dot)
+   bfd_vma dot,
+   bfd_size_type *alignment_needed)
 {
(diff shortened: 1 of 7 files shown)


Nelson1225:
Thanks for the report.

It is better to resolve the specific target issue in the target related files rather than the generic code.  If you modify the generic code, then you have to make sure that your modification won't affect other ports. 

I suppose that the default linker script won't cause this problem.  See the `_bfd_riscv_relax_call` in the bfd/elfnn-riscv.c, we have already consider the maximum alignment for the cross output section boundaries.  Your failed case will gather the different input "text" sections with different alignments into the same output section.  I'm not sure whether this is a normal usage, but if we need to resolve this, then always consider the max alignment for the call relaxation may be a more suitable way.

Thanks
Best Regards
Nelson

yiting16:
@Nelson1225 Thank you for your guidance.  I understand what you mean. Jim has created a patch, used the max alignment for the call relaxation as you said. So, I will close my pull request.

PR #187, opened 2019-11-19
Riscv binutils 2.33.1 rvv 0.8.x spec update

Nelson1225:
Update to the version 0.8-draft-20191117.  The constraints for all RVV instructions, including the new instructions, will be updated in the later patches.

1. Add vaaddu.v[v|x] and vasubu.v[v|x], change the encoding of vaadd.v[v|x] and vasub.v[v|x].  Remove the vaadd.vi.

    * vaaddu.v[v|x]: New, OPMV[V|X], funct6 is 0x001000.
    * vaadd.v[v|x]: Change from OPIV[V|X] to OPMV[V|X], change funct6 from 0x100100 to 0x001001.
    * vaadd.vi: Removed.
    * vasubu.v[v|x]: New, OPMV[V|X], funct6 is 0x001010.
    * vasub.v[v|x]: Change from OPIV[V|X] to OPMV[V|X], change funct6 from 0x100110 to 0x001011.

2. Remove scaled fixed-point multiply-add.

    vwsmaccu.v[v|x], vwsmacc.v[v|x], vwsmaccus.vx and vwsmaccsu.v[v|x].

3. Added tentative quad-widening multiply-add instructions.  The encoding of the tentative quad-widening multiply-add instructions are the same as the removed scaled fixed-point multiply-add.

    vqmaccu.v[v|x], vqmacc.v[v|x], vqmaccus.vx and vqmaccsu.v[v|x].

4. Added support for whole vector register moves.

    Load/stores of multiple whole vector registers (ignoring vl and vtype) and
    multiple whole vector register move instructions.  Also, new vlenb CSR to
    report number of bytes required to hold a single vector register.

    The encoding of vl<nf>r.v is similar as unmasked zero-extended unit-stride
    load, and the encoding of vs<nf>r.v is similar as unmasked unit-stride store.
    The vmv<nf>r.v is encoded as an OPIVI instruction, and uses the same funct6
    encoding as the `vsmul` instruction but with an immediate operand.


jim-wilson:
FYI About the "temporary" comments in include/opcode/riscv-opc.h, I wasn't planning on submitting those upstream.  They exist to help me track the encoding changes from one draft version to the next.  The encodings are still changing, so they are still useful.  But I'm thinking we should strip those comments out before we eventually send upstream.

jim-wilson:
Aren't we missing a change for vmadc/vmsbc?  They have optional mask/carry-in inputs now.  I only see support for the masked versions.  Support for the unmasked versions seems to be missing.  This can be submitted as a separate patch.

Nelson1225:
> FYI About the "temporary" comments in include/opcode/riscv-opc.h, I wasn't planning on submitting those upstream. They exist to help me track the encoding changes from one draft version to the next. The encodings are still changing, so they are still useful. But I'm thinking we should strip those comments out before we eventually send upstream.

Yes, the version tag comments just be used to help us track the change between spec versions.  Thanks for the reminder that we should strip those tag comments out before sending upstream :)

Nelson1225:
> Aren't we missing a change for vmadc/vmsbc? They have optional mask/carry-in inputs now. I only see support for the masked versions. Support for the unmasked versions seems to be missing. This can be submitted as a separate patch.

I remember that this change is committed by another pull request, and we have already supported it on the riscv-binutils-2.33.1-rvv-0.8.x branch now :)

https://github.com/riscv/riscv-binutils-gdb/commit/c44ea79ca538eba4713fdf6c454a59009c7d5e1c

jim-wilson:
The riscv-gnu-toolchain rvv-0.8.x branch is missing the last two commits to the riscv-binutils rvv-0.8.x branch.  I forgot to check for that.  We need to remember to update riscv-gnu-toolchain when riscv-binutils-gdb is updated.

kito-cheng:
@Nelson1225 could you help to bump the `rvv-0.8.x branch` for `riscv-gnu-toolchain`, let me know if you don't know how to do.

Nelson1225:
Sure, I will remember bumping the branch for `riscv-gnu-toolchain` once we update the  riscv-binutils-gdb.  Thanks for the reminder :)

PR #188, opened 2019-11-22
Riscv binutils 2.33.1 rvv 0.8.x constraint

Nelson1225:
Currently, assembler is hard to know what the LMUL exactly is when assembling.
Therefore, we can just check the part of the constraints in the static assemble-time.

The currently supported vector constraints:
* Widening and narrowing vector arithmetic instructions
* Quad-widening integer multiply-add
* Vector indexed segment loads and stores
* Vector Iota
* Vector slideup and slide1up
* Vector register gather and compress
* Vector whole register load, store and move

Another thing is that we just report "Error: illegal operands" messages for all
vector constraints now.  It would be nice if we can report the clearer error
message, so this may be the next stuff we can improve.

Thanks
Best Regards
Nelson


kito-cheng:
Does it possible to provide more clear error message than just all `Error: illegal operands ` ?

kito-cheng:
Oh, I see your comment

Nelson1225:
> Oh, I see your comment

Yes, and I think we can improve it by extend the "insn->match_func" in the riscv_ip.

For example,
insn->match_func (insn, ip->insn_opcode, **&error**);

and then we can update different error message by different `match_func`.

But we need to check if other assembler changes are required.  For now, when assembler parsing one assembly instruction, we will check all candidates which have the same opcode name.  So stop the checking and report error message before the last candidate shouldn't be an expected behavior.  Also, if there are too many candidates, then it is hard to report the correct error message, like the ".insn" directive.  It would be nice if we can improve that :)

Nelson1225:
The series commits should be updated based on the following pull request,
https://github.com/riscv/riscv-binutils-gdb/pull/191

Therefore, I will fix the conflicts and update the commits after the above pull request is approved :)

Thanks and best regards
Nelson

Nelson1225:
Updated.

* No need to rename the gas/testsuite/gas/riscv/vector-insns to gas/testsuite/gas/riscv/vector-insns-ok.
* Vector slide up and slide 1 up don't allow vd overlap vs2 and vm.
* Only support single vector register moved for loads and stores (vl1r and vs2r), so don't need to do the contraint checking for them.
*  Only support the standard aligned vector register groups (vmv1r, vmv2r, vmv4r, vmv8r) currently, so just check the vd and vs2 are aligned to nf.

Nelson1225:
Thanks for the approval :)
I will start to improve the error message recently.

PR #190, opened 2019-12-03
Update the operand's order of VAMO instructions.

Nelson1225:
Change the format from

vamo vd, v2, (rs1), vd, v0.t

to 

vamo vd, (rs1), v2, vd, v0.t


Nelson1225:
Yes I notice that the spec update the syntax of vamo after the implementation in binutils.  And I also miss this when I update the rvv-0.7.x to rvv-0.8.x.  Thanks for the quick approval :)

PR #191, opened 2019-12-12
Riscv binutils 2.33.1 rvv 0.8.x spec 20191207

Nelson1225:
Only single vector register moved for loads and stores, and only standard aligned vector register groups are allowed.  The remaining encode are reserved to allow future multiple load/stores and more flexible register-register moves.  This won't change the encoding of the current vector instructions.

The RVV constraints are also need to be updated.  I will update them in the Nelson1225:riscv-binutils-2.33.1-rvv-0.8.x-constraint branch.


Nelson1225:
Thanks a lot.

PR #192, opened 2019-12-16
Accept constant 0 for address of vector amo instruction

kito-cheng:
Scalar AMO instruction and other vector load store can accept address start with 0, so I think it would be better to support that for vector AMO for symmetric.


Nelson1225:
Agree with you :)

PR #193, opened 2019-12-24
Add GAS option to disable the instruction constraints checking.

Nelson1225:
I get the request recently that the assembler's constraints checking break hardware exception test cases.  As I know, stop to assemble instructions with constraints is reasonable to me.  But it is hard to check whether the hardware  exceptions work expected for the constraints.  For now, the compromises are use the hard cored instruction (.word or .2byte) and .insn directives to build the test cases.  But I think the current test cases won't built by them and can not check the exception for constraints.

It seems quite easy to resolve the problem by adding a GAS internal option to disable checking the RVV constraints (always call `match_opcde` to check).  But for the normal ISA (scalar), we have much complex parsing method for the constraints (The same opcode name with different operands, Co and Cj, ...).  It doesn't make sense to me that we can use a internal option to affect the current operand parsing.  Also, add a internal option just used to disable the RVV constraints is hard to maintain.

My personal opinion is not to support a internal option used to disable the constraints.  But I can not convince them to use the hard-cored instruction and .insn directive to build the test cases in the sort-term.  Since the former is hard to be generated automatically, and the later for RVV isn't complete now (the RVV instruction name are still discussed).

Ideally, I suppose our assembler shouldn't allow to generate the instruction with constraints.  The internal usage option is also hard to maintain.  It would be great if the constraint test cases can be built by the .insn directives.  I hope we can have the leeway that not to support the internal usage option, which just for testing, in the long-term.  But in the sort-term,  it is understandable that the internal option is easy to develop (build the test cases).

### DIFF
--- a/gas/config/tc-riscv.c
+++ b/gas/config/tc-riscv.c
@@ -83,6 +83,7 @@ struct riscv_set_options
   int rve; /* Generate RVE code.  */
   int relax; /* Emit relocs the linker is allowed to relax.  */
   int arch_attr; /* Emit arch attribute.  */
+  int check_constraints; /* Enable/disable the match_func checking.  */
 };
 
 static struct riscv_set_options riscv_opts =
@@ -92,6 +93,7 @@ static struct riscv_set_options riscv_opts =
   0,	/* rve */
   1,	/* relax */
   DEFAULT_RISCV_ATTR, /* arch_attr */
+  0,	/* check_constraints */
 };
 
 static void
@@ -1648,7 +1650,8 @@ riscv_ip (char *str, struct riscv_cl_insn *ip, expressionS *imm_expr,
 	    case '\0': 	/* End of args.  */
 	      if (insn->pinfo != INSN_MACRO)
 		{
-		  if (!insn->match_func (insn, ip->insn_opcode))
+		  if (!insn->match_func (insn, ip->insn_opcode,
+					 riscv_opts.check_constraints))
 		    break;
 
 		  /* For .insn, insn->match and insn->mask are 0.  */
@@ -2531,6 +2534,8 @@ enum options
   OPTION_NO_RELAX,
   OPTION_ARCH_ATTR,
   OPTION_NO_ARCH_ATTR,
+  OPTION_CHECK_CONSTRAINTS,
+  OPTION_NO_CHECK_CONSTRAINTS,
   OPTION_END_OF_ENUM
 };
 
@@ -2545,6 +2550,8 @@ struct option md_longopts[] =
   {"mno-relax", no_argument, NULL, OPTION_NO_RELAX},
   {"march-attr", no_argument, NULL, OPTION_ARCH_ATTR},
   {"mno-arch-attr", no_argument, NULL, OPTION_NO_ARCH_ATTR},
+  {"mcheck-constraints", no_argument, NULL, OPTION_CHECK_CONSTRAINTS},
+  {"mno-check-constraints", no_argument, NULL, OPTION_NO_CHECK_CONSTRAINTS},
 
   {NULL, no_argument, NULL, 0}
 };
@@ -2621,6 +2628,14 @@ md_parse_option (int c, const char *arg)
 
     case OPTION_NO_ARCH_ATTR:
       riscv_opts.arch_attr = FALSE;
+      break;
+
+    case OPTION_CHECK_CONSTRAINTS:
+      riscv_opts.check_constraints = TRUE;
+      break;
+
+    case OPTION_NO_CHECK_CONSTRAINTS:
+      riscv_opts.check_constraints = FALSE;
       break;
 
     default:
@@ -3009,6 +3024,10 @@ s_riscv_option (int x ATTRIBUTE_UNUSED)
     riscv_opts.relax = TRUE;
   else if (strcmp (name, "norelax") == 0)
     riscv_opts.relax = FALSE;
+  else if (strcmp (name, "checkconstraints") == 0)
+    riscv_opts.check_constraints = TRUE;
+  else if (strcmp (name, "nocheckconstraints") == 0)
+    riscv_opts.check_constraints = FALSE;
   else if (strcmp (name, "push") == 0)
     {
       struct riscv_option_stack *s;
--- a/include/opcode/riscv.h
+++ b/include/opcode/riscv.h
@@ -395,8 +395,11 @@ struct riscv_opcode
      INSN_MACRO, then this field is the macro identifier.  */
   insn_t mask;
   /* A function to determine if a word corresponds to this instruction.
-     Usually, this computes ((word & mask) == match).  */
-  int (*match_func) (const struct riscv_opcode *op, insn_t word);
+     Usually, this computes ((word & mask) == match).  If the constraints
+     checking is disable, then most of the function should check only the
+     basic encoding for the instruction.  */
+  int (*match_func) (const struct riscv_opcode *op, insn_t word,
+		     int constraints);
   /* For a macro, this is INSN_MACRO.  Otherwise, it is a collection
      of bits describing the instruction, notably any relevant hazard
      information.  */
--- a/include/opcode/riscv.h
+++ b/include/opcode/riscv.h
@@ -24,6 +24,7 @@
 #include "riscv-opc.h"
 #include <stdlib.h>
 #include <stdint.h>
+#include "bfd.h"
 
 typedef uint64_t insn_t;
 
@@ -398,8 +399,8 @@ struct riscv_opcode
      Usually, this computes ((word & mask) == match).  If the constraints
      checking is disable, then most of the function should check only the
      basic encoding for the instruction.  */
-  int (*match_func) (const struct riscv_opcode *op, insn_t word,
-		     int constraints);
+  bfd_boolean (*match_func) (const struct riscv_opcode *op, insn_t word,
+			     bfd_boolean constraints);
   /* For a macro, this is INSN_MACRO.  Otherwise, it is a collection
      of bits describing the instruction, notably any relevant hazard
      information.  */
--- a/opcodes/riscv-dis.c
+++ b/opcodes/riscv-dis.c
@@ -507,7 +507,7 @@ riscv_disassemble_insn (bfd_vma memaddr, insn_t word, disassemble_info *info)
       for (; op->name; op++)
 	{
 	  /* Does the opcode match?  */
-	  if (! (op->match_func) (op, word))
+	  if (! (op->match_func) (op, word, 0))
 	    continue;
 	  /* Is this a pseudo-instruction and may we print it as such?  */
 	  if (no_aliases && (op->pinfo & INSN_ALIAS))
--- a/opcodes/riscv-dis.c
+++ b/opcodes/riscv-dis.c
@@ -507,7 +507,7 @@ riscv_disassemble_insn (bfd_vma memaddr, insn_t word, disassemble_info *info)
       for (; op->name; op++)
 	{
 	  /* Does the opcode match?  */
-	  if (! (op->match_func) (op, word, 0))
+	  if (! (op->match_func) (op, word, FALSE))
 	    continue;
 	  /* Is this a pseudo-instruction and may we print it as such?  */
 	  if (no_aliases && (op->pinfo & INSN_ALIAS))
--- a/opcodes/riscv-opc.c
+++ b/opcodes/riscv-opc.c
@@ -98,83 +98,106 @@ const char * const riscv_vecm_names_numeric[NVECM] =
 #define MASK_VMASK (OP_MASK_VMASK << OP_SH_VMASK)
 
 static int
-match_opcode (const struct riscv_opcode *op, insn_t insn)
+match_opcode (const struct riscv_opcode *op,
+	      insn_t insn,
+	      int constraints ATTRIBUTE_UNUSED)
 {
   return ((insn ^ op->match) & op->mask) == 0;
 }
 
 static int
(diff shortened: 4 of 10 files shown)


jim-wilson:
In another thread, t was Palmer and/or Andrew that mentioned that adding constraint checking to the assembler might break hardware verification tests.  I think there are a few things we can do about this.  We could have the compiler add the new option by default, so that people using the compiler get the errors by default, but people writing assembly code don't.  Or alternatively we can add a .option check (or whatever) to enable the checking which the compiler emits by default.  I think this might be better as it is more consistent with the current approach for other stuff like rvc/norvc and pic/nopic.  It also lets you turn the checking on/off in the middle of a file if necessary, which might be useful if someone needs an invalid instruction, but wants the rest of it to be valid code.

gas/config/tc-riscv.c
```diff
@@ -2546,6 +2554,8 @@ struct option md_longopts[] =
   {"march-attr", no_argument, NULL, OPTION_ARCH_ATTR},
   {"mno-arch-attr", no_argument, NULL, OPTION_NO_ARCH_ATTR},
 
+  /* These options are used for internal usage or testing.  */
+  {"mno-check-constraints", no_argument, NULL, OPTION_NO_CHECK_CONSTRAINTS},
```
jim-wilson: It is better to add options in pairs.  Someone will always find a reason to use the other one.  For instance, if you assemble with -mno-check-constraints, and then want to assemble again with constraints checks enabled, then you can just use "!! -mcheck-constraints", but if you don't have the positive option you have to edit the last command which takes more time then just adding another option at the end.

aswaterman:
Something like a .option that the compiler emits by default seems pretty OK to me. Breaking the current tests does not seem acceptable, not just because of the extra support issues that will raise, but also because other people have mimicked the style of those tests, so the problem is more widespread than it might seem.

gas/config/tc-riscv.c
Nelson1225: Agree with this, thanks :)

Nelson1225:
@jim-wilson @aswaterman, Thank you very much.  This issue are worse than I thought.  I have sent a series of CSR checking commits to upstream, and also meets the same problem as here.  Adding GAS options and .option directives to enable/disable these checking is a good solution.  We disable these checking in assembler by default, and then enable them by compiler and user adding the new options.  It is reasonable to me.

I plan to add both GAS option and .option to control the constraints checking, like the CSR checking.  If the constraints checking is disabled, then we always call `match_opcode` rather than the specific `match_func` to check only the basic encoding.  For now I'm going to improve the assembler's error reporting.  So after that, we will see what more these constraints options can do :)

Nelson1225:
Ideally, the `match_func` should only need to check the basic encoding if the checking is disabled.  But there are several cases may break this rules, since not all `match_func` are only used to check the instruction constraints.

1. `match_vs1_eq_vs2` and `match_vd_eq_vs1_eq_vs2` with `INSN_ALIAS`.
These are used for disassembler.  `vmcpy.m` and `vmand.mm` have the same match and mask macros (MATCH_VMANDMM and MASK_VMANDMM), but use the `match_vs1_eq_vs2` to decide which one to disassemble.

2. `match_c_add`, `match_c_lui`, `match_slli_as_c_slli` and `match_srxi_as_c_srxi`.
These are used to avoid converting the RVI instructions to the corresponding hint RVC.

However, I prefer not to change the original behavior of non-vector instructions in this commit and rvv branch.  Therefore, the new constraints options can only affect the following `match_func` so far,

`match_widen_vd_neq_vs1_neq_vs2_neq_vm`,
`match_widen_vd_neq_vs1_neq_vm`,
`match_widen_vd_neq_vs2_neq_vm`,
`match_widen_vd_neq_vm`,
`match_quad_vd_neq_vs1_neq_vs2_neq_vm`,
`match_quad_vd_neq_vs2_neq_vm`,
`match_narrow_vd_neq_vs2`,
`match_vd_neq_vs1_neq_vs2_neq_vm`,
`match_vd_neq_vs2_neq_vm`,
`match_vd_neq_vm`,
`match_vmv_nf_rv`

Beside, I think use new options to control the `match_func` of `vwcvt.x.x.v` and `vwcvtu.x.x.v` with INSN_ALIAS is safe.  Since their corresponding non-alias instructions (vwadd.vx and vwaddu.vx) are also use the same `match_func` to check the constraints (`match_widen_vd_neq_vs2_neq_vm`), and have different match and mask macros.

Nelson1225:
Unified the logic of match_widen* and match_narrow* functions.  Fixed some typo.

include/opcode/riscv.h
```diff
@@ -395,8 +395,11 @@ struct riscv_opcode
      INSN_MACRO, then this field is the macro identifier.  */
   insn_t mask;
   /* A function to determine if a word corresponds to this instruction.
-     Usually, this computes ((word & mask) == match).  */
-  int (*match_func) (const struct riscv_opcode *op, insn_t word);
+     Usually, this computes ((word & mask) == match).  If the constraints
+     checking is disable, then most of the function should check only the
+     basic encoding for the instruction.  */
+  int (*match_func) (const struct riscv_opcode *op, insn_t word,
+		     int contraints);
```
kito-cheng: typo, `contraints`->`constraints`, and `bfd_boolean` for constraints would be better.

opcodes/riscv-dis.c
```diff
@@ -507,7 +507,7 @@ riscv_disassemble_insn (bfd_vma memaddr, insn_t word, disassemble_info *info)
       for (; op->name; op++)
 	{
 	  /* Does the opcode match?  */
-	  if (! (op->match_func) (op, word))
+	  if (! (op->match_func) (op, word, 0))
```
kito-cheng: `FALSE` instead `0`

opcodes/riscv-opc.c
```diff
@@ -98,83 +98,106 @@ const char * const riscv_vecm_names_numeric[NVECM] =
...
 static int
-match_rs1_eq_rs2 (const struct riscv_opcode *op, insn_t insn)
+match_rs1_eq_rs2 (const struct riscv_opcode *op,
+		  insn_t insn,
+		  int constraints ATTRIBUTE_UNUSED)
 {
   int rs1 = (insn & MASK_RS1) >> OP_SH_RS1;
   int rs2 = (insn & MASK_RS2) >> OP_SH_RS2;
-  return match_opcode (op, insn) && rs1 == rs2;
+  return match_opcode (op, insn, 0) && rs1 == rs2;
```
kito-cheng: `FALSE` instead `0`

opcodes/riscv-opc.c
```diff
@@ -98,83 +98,106 @@ const char * const riscv_vecm_names_numeric[NVECM] =
...
+match_vd_eq_vs1_eq_vs2 (const struct riscv_opcode *op,
+			insn_t insn,
+			int constraints ATTRIBUTE_UNUSED)
 {
   int vd =  (insn & MASK_VD) >> OP_SH_VD;
   int vs1 = (insn & MASK_VS1) >> OP_SH_VS1;
   int vs2 = (insn & MASK_VS2) >> OP_SH_VS2;
-  return match_opcode (op, insn) && vd == vs1 && vs1 == vs2;
+
+  return match_opcode (op, insn, 0) && vd == vs1 && vs1 == vs2;
```
kito-cheng: Ditto.

include/opcode/riscv.h
Nelson1225: Fixed

opcodes/riscv-dis.c
Nelson1225: fixed

opcodes/riscv-opc.c
Nelson1225: fixed

opcodes/riscv-opc.c
Nelson1225: fixed

Nelson1225:
Thanks you very much @kito-cheng :)
Fixed the typo.  Changed the return type of `match_func` and `constraints` flag to `bfd_boolean`.

opcodes/riscv-opc.c
```diff
@@ -97,291 +97,370 @@ const char * const riscv_vecm_names_numeric[NVECM] =
...
 
-static int
-match_c_lui (const struct riscv_opcode *op, insn_t insn)
+static bfd_boolean
+match_c_lui (const struct riscv_opcode *op,
+	     insn_t insn,
+	     bfd_boolean constraints ATTRIBUTE_UNUSED)
 {
-  return (match_rd_nonzero (op, insn)
+  return (match_rd_nonzero (op, insn, 0)
```
kito-cheng: `FALSE` instead of `0`

opcodes/riscv-opc.c
```diff
@@ -97,291 +97,370 @@ const char * const riscv_vecm_names_numeric[NVECM] =
...
 
-static int
-match_slli_as_c_slli (const struct riscv_opcode *op, insn_t insn)
+static bfd_boolean
+match_slli_as_c_slli (const struct riscv_opcode *op,
+		      insn_t insn,
+		      bfd_boolean constraints ATTRIBUTE_UNUSED)
 {
-  return match_rd_nonzero (op, insn) && EXTRACT_RVC_IMM (insn) != 0;
+  return match_rd_nonzero (op, insn, 0) && EXTRACT_RVC_IMM (insn) != 0;
```
kito-cheng: Ditto.

opcodes/riscv-opc.c
```diff
@@ -97,291 +97,370 @@ const char * const riscv_vecm_names_numeric[NVECM] =
...
-	  && (vd % 2) == 0
-	  && (vs1 < vd || vs1 > (vd + 1))
-	  && (vs2 < vd || vs2 > (vd + 1))
-	  && (vm || vm < vd || vm > (vd + 1)));
+  if (constraints
+      && ((vd % 2) != 0
+	  || (vs1 >= vd && vs1 <= (vd + 1))
+	  || (vs2 >= vd && vs2 <= (vd + 1))
+	  || (!vm && vm >= vd && vm <= (vd + 1))))
+    return 0;
```
kito-cheng: Ditto.

opcodes/riscv-opc.c
```diff
@@ -97,291 +97,370 @@ const char * const riscv_vecm_names_numeric[NVECM] =
...
-	  && (vd % 2) == 0
-	  && (vs2 % 2) == 0
-	  && (vs1 < vd || vs1 > (vd + 1))
-	  && (vm || vm < vd || vm > (vd + 1)));
+  if (constraints
+      && ((vd % 2) != 0
+	  || (vs2 % 2) != 0
+	  || (vs1 >= vd && vs1 <= (vd + 1))
+	  || (!vm && vm >= vd && vm <= (vd + 1))))
+    return 0;
```
kito-cheng: Ditto.

opcodes/riscv-opc.c
```diff
@@ -97,291 +97,370 @@ const char * const riscv_vecm_names_numeric[NVECM] =
...
 
-  return (match_opcode (op, insn)
-	  && (vd % 2) == 0
-	  && (vs2 < vd || vs2 > (vd + 1))
-	  && (vm || vm < vd || vm > (vd + 1)));
+  if (constraints
+      && ((vd % 2) != 0
+	  || (vs2 >= vd && vs2 <= (vd + 1))
+	  || (!vm && vm >= vd && vm <= (vd + 1))))
+    return 0;
```
kito-cheng: Ditto.

opcodes/riscv-opc.c
```diff
@@ -97,291 +97,370 @@ const char * const riscv_vecm_names_numeric[NVECM] =
...
 
-  return (match_opcode (op, insn)
-	  && (vd % 2) == 0
-	  && (vs2 % 2) == 0
-	  && (vm || vm < vd || vm > (vd + 1)));
+  if (constraints
+      && ((vd % 2) != 0
+	  || (vs2 % 2) != 0
+	  || (!vm && vm >= vd && vm <= (vd + 1))))
+    return 0;
```
kito-cheng: Ditto.

opcodes/riscv-opc.c
Nelson1225: Sorry for the neglect.  Fixed.

opcodes/riscv-opc.c
Nelson1225: Fixed.

opcodes/riscv-opc.c
Nelson1225: Fixed.

opcodes/riscv-opc.c
Nelson1225: Fixed.

opcodes/riscv-opc.c
Nelson1225: Fixed.

opcodes/riscv-opc.c
Nelson1225: Fixed.

Nelson1225:
The related support in GCC https://github.com/riscv/riscv-gcc/pull/173

Nelson1225:
Thank you very much, Jim and Kito.  Thanks for the review and approval :)

Nelson1225:
> I think we painted ourselves in a corner by having for instance norelax instead of no-relax at the beginning. 

Yeah, the naming of command line options are fine, but the naming of .option directive are not consistent with the former...  It would be good if we have the consistent naming from now, and have some remedies like you said, always add extra .option names later though :)

aswaterman:
The `norelax` is inherited from the MIPS option name `noreorder`.  Not the best decision on my part.  I guess we could start using hyphens in new names, and for the legacy ones, we can create aliases with hyphens.

PR #197, opened 2020-02-10
Support RISC-V ZV* extensions in riscv-binutils-2.33.1-rvv-0.8.x branch

Nelson1225:
Support RISC-V ZV* extensions.  There are four Z sub-extensions for vector, Zvamo, Zvediv, Zvlsseg and Zvqmac.  The vector base is present before adding any of current Zv* extensions.

* Zvamo (vamo*)
Enable vector AMO instructions.  If the extension is supported, then the atomic
operations from the standard A extension must be present, too.

* Zvediv
If the extension isn't supported, then the vediv field of vtype shouldn't be
set by vsetvli instruction.  Report the warning message for the constraint case.

* Zvlsseg (vlseg*, vlsseg*, vlxseg, vsseg*, vssseg* and vsxseg*)
Enable vector load/store segment instructions.

* Zvqmac (vqmaccu*, vqmaccsu* and vqmaccus*)
Enable vector quad-widening integer multiply-add instructions.


kito-cheng:
I would like put those patch based on binutils 2.34 and rvv-0.9.x branch, so I gonna close this PR.

Nelson1225:
> I would like put those patch based on binutils 2.34 and rvv-0.9.x branch, so I gonna close this PR.

OK, noted :)

PR #199, opened 2020-02-21
Update rvv implementation from riscv-binutils-2.33.1-rvv-0.8.x to rvv-0.9.x

Nelson1225:
rvv-0.9.x branch is based on riscv-binutils-2.34 branch.  We have some fixed bugs and some good features in the binutils-2.34-release branch, so it would be great if the rvv branch also have these good features.  Beside, update binutils for rvv can reduce the burden of contributing to upstream.


jim-wilson:
I noticed that rvv-0.8.x has the binutils/gcc csr-checking patches, but rvv-0.9.x does not.  This pull request adds the missing binutils csr-checking patch.  You still need to add the missing gcc csr-checking patch.

Nelson1225:
OK, thanks for the friendly reminder.  I will bump riscv-gcc and riscv-binutils to support the checking in rvv-0.9.x.

PR #200, opened 2020-02-27
Support Zv* sub extensions in rvv-0.9.x branch

Nelson1225:
There are four Z sub-extensions for vector, Zvamo, Zvediv, Zvlsseg and Zvqmac.  The vector base is present before adding any of current Zv* extensions.

* Zvamo (vamo*)
Enable vector AMO instructions.  If the extension is supported, then the atomic operations from the standard A extension must be present, too.

* Zvediv
If the extension isn't supported, then the vediv field of vtype shouldn't be set by vsetvli instruction.  Report the warning message for the constraint case.  Since we just report the warning message for the constraint case, we still regard vetvli as an INSN_CLASS_V instruction rather than an INSN_CLASS_V_AND_ZVEDIV.  We check whether the zvediv ext is enabled when parsing the Vc operand.

* Zvlsseg (vlseg*, vlsseg*, vlxseg, vsseg*, vssseg* and vsxseg*)
Enable vector load/store segment instructions.

* Zvqmac (vqmaccu*, vqmaccsu* and vqmaccus*)
Enable vector quad-widening integer multiply-add instructions.


gas/config/tc-riscv.c
```diff
@@ -2388,6 +2398,11 @@ riscv_ip (char *str, struct riscv_cl_insn *ip, expressionS *imm_expr,
 		  if (!VALID_RVV_VC_IMM (imm_expr->X_add_number))
 		    as_bad (_("bad value for vsetvli immediate field, "
 			      "value must be 0..2047"));
+		  /* Report warning message if the vediv field is set, but the
+		     Zvediv extension isn't enabled.  */
+		  if (!riscv_multi_subset_supports (INSN_CLASS_V_AND_ZVEDIV)
+		      && (imm_expr->X_add_number & (0x3 << 5)))
```
jim-wilson: We have macros for these magic constants, OP_MASK_VEDIV and OP_SH_VEDIV, which would be better to use.  Though I see that my_getVsetvliExpression isn't using the OP_SH_* macros either and should be fixed, that is probably my fault.

It looks a little odd that my_getVsetvliExpression accepts 'd' without checking INSN_CLASS_V_AND_ZVEDIV.  But you do catch it here.  Maybe add a comment that the caller is expected to check this.  Or maybe move the check into that function.  Shrug.  I don't think it matters.

jim-wilson:
I'm assuming two patches cherry picked from the FSF tree, and I only need to review the one new one.

gas/config/tc-riscv.c
Nelson1225: > We have macros for these magic constants, OP_MASK_VEDIV and OP_SH_VEDIV, which 
> would be better to use. Though I see that my_getVsetvliExpression isn't using the OP_SH_* 
> macros either and should be fixed

Yeah it is better to use OP_* macros, I will fix this, and also fix the my_getVsetvliExpression in another commit.

Nelson1225: > It looks a little odd that my_getVsetvliExpression accepts 'd' without checking 
 INSN_CLASS_V_AND_ZVEDIV. But you do catch it here. Maybe add a comment that the  caller is expected to check this. Or maybe move the check into that function. Shrug. I don't think it matters.

The reason for the beginning is that I want to check the vediv extension after the bad value checking for immediate settings (vsetvli t0, a0, 0x2047).  Hmm, but it seems good to move the checking into the function, and the order of two checking (bad value checking for imm setting and vediv checking) shouldn't matters.

Nelson1225:
All fixed.  The first and second commits are picked from upstream.  I move the zvediv extension checking into the my_getVsetvliExpression in the third commit. And replace some magic constants with the OP_* masks for my_getVsetvliExpression in the forth commit.

Thanks a lot :)

PR #201, opened 2020-02-27
Support vector CSR checking in rvv-0.9.x branch

Nelson1225:
The first, third and fourth patches are picked from upstream binutils.  The second patch "RISC-V: Support v-ext for the ISA-dependent CSR checking" is the new one.


jim-wilson:
I'm assuming 3 patches cherry picked from the FSF tree and I only need to review the one new one.

Nelson1225:
Thanks :)

Issue #206, opened 2020-04-01
Strange interpret of pseudo instruction vmsltu.vi


HanKuanChen:
I am using `rvv-0.9.x` branch.

I have a `.S` which contains `vmsltu.vi v25, v14, 0`.
`v25` should be `0` for all elements.
But the assembler interprets this instruction into `vmsleu.vi v25,v14,-1`, which makes `v25` be `1` for all elements.

aswaterman:
Options are:

- Reject the pseudoinstruction when imm=0.
- Convert the pseudoinstruction to `vmsne.vv v25, v14, v14` when imm=0, which has the correct effect of setting all the elements to 0 (and also it maintains any illegal-instruction trapping behavior because of the register numbers).

kito-cheng:
> Options are:
> 
> * Reject the pseudoinstruction when imm=0.
> * Convert the pseudoinstruction to `vmsne.vv v25, v14, v14` when imm=0, which has the correct effect of setting all the elements to 0 (and also it maintains any illegal-instruction trapping behavior because of the register numbers).

I prefer second option, because it's pseudoinstruction, it possible to make it by another instruction, so why not?

@Nelson1225

Nelson1225:
Should we convert the vmsgeu.vi with zero imm to vmsne.vv too?

aswaterman:
The spec says "Note, vmsgeu.vi with immediate 0 is not useful as it is always true" -- I think Krste was implying that "there's no reason to provide a pseudo-op, because the instruction isn't useful."

I don't feel strongly one way or the other, but we should be consistent between vmsltu and vmsgeu (make both illegal with 0 operation, or convert both to vmsne/vmseq)

Nelson1225:
> The spec says "Note, vmsgeu.vi with immediate 0 is not useful as it is always true" -- I think Krste was implying that "there's no reason to provide a pseudo-op, because the instruction isn't useful."
>
> I don't feel strongly one way or the other, but we should be consistent between vmsltu and vmsgeu (make both illegal with 0 operand, or convert both to vmsne/vmseq)

Agree, two method are good to me.  I will create a riscv-v-spec issue to make sure which way is preferred by team :)

Thanks

Nelson1225:
@HanKuanChen, This should be fixed in the rvv-0.9.x branch.  Thanks :)

PR #207, opened 2020-04-07
Change the expansions and conversion for pseudo vector compare instructions (integer and scale).

Nelson1225:
There are two issues about the vector compare instructions recently,

1. Convert the pseudo vmsltu.vi/vmsgeu.vi to vmsne.vv/vmseq.vv, when imm is zero. 
(https://github.com/riscv/riscv-binutils-gdb/issues/206)

We convert the pseudo instructions

	`vmsltu.vi/vmsgeu.vi  vd, vs1, imm`

to

	(imm isn't zero) `vmsleu.vi/vmsgtu.vi  vd, vs1, imm - 1`
	(imm is zero)    `vmsne.vv/vmseq.vv    vd, vs1, vs1`

2. Change the expansion for vmsge.vx and vmsgeu.vx with the unmasked va.
(https://github.com/riscv/riscv-binutils-gdb/issues/208)

`vmsge[u].vx v4, v8, a1` should be expanded to `vmslt[u].vx  v4,v8,a1` plus `vmnot.m  v4,v4`.

### DIFF
--- a/gas/config/tc-riscv.c
+++ b/gas/config/tc-riscv.c
@@ -1279,7 +1279,7 @@ vector_macro (struct riscv_cl_insn *ip)
 	{
 	  /* Unmasked.  */
 	  macro_build (NULL, "vmslt.vx", "Vd,Vt,sVm", vd, vs2, vs1, -1);
-	  macro_build (NULL, "vmnand.mm", "Vd,Vt,Vs", vd, vs2, vs1);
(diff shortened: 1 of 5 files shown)


Nelson1225:
> Looks like two separate fixes, only one of which is mentioned in the explanation, but both look correct.

Yes, these are two separate issues.  The first one is single instruction conversion, and the second one is expansion.  I must miss the description for the first one...

kito-cheng:
@Nelson1225 Could you back port this two fix into rvv-0.8.x bracnh?

PR #212, opened 2020-05-09
rvv-0.9.x fix typo of macro

katsuster:
This patch fixes typo in macro name of MASK_ and MATCH_ of riscv
instructions.


jim-wilson:
Contributing to FSF projects like binutils requires a copyright assignment.  If we accept patches from people without assignments, we will not be able to contribute the patches upstream to the FSF when the vector extension becomes official.

Do you have a copyright assignment?  I don't see one registered for you at the FSF, but I have only your name, and don't know your employer.  Are you willing to get an assignment?  Whether you can get one depends on where you work, as some companies are willing to go through the process and some are not.  Even if your employer is willing to do the work, it may take a few months to complete the paperwork.

Note that the simple solution is to just file a bug report, and let someone with an assignment write the necessary patches.

katsuster:
Thank you for your comments.
I don't have copyright assignment yet. I will try to get it.

katsuster:
Sorry for late.
Today, I have received copyright assignment of GCC and binutils from FSF.
(FSF's accepted date is 2020 May 29)
It is no problem to merge my patch for binutils.

PR #213, opened 2020-05-09
Fix typo of mask and match macro name of riscv

katsuster:
This patch fixes typo in macro name of MASK_ and MATCH_ of riscv
instructions.


jim-wilson:
Same as #212 except for the 0.8 branch instead of the 0.9 branch.  Asked about copyright assignments in 212.

PR #219, opened 2020-06-05
The Zvamo and Zvlsseg should be included in the base V extension.

Nelson1225:
According to the following description in vector spec,
https://github.com/riscv/riscv-v-spec/blob/master/v-spec.adoc#8-vector-amo-operations-zvamo
https://github.com/riscv/riscv-v-spec/blob/master/v-spec.adoc#78-vector-loadstore-segment-instructions-zvlsseg

The Zvamo and Zvlsseg are intended to be included in the base V extension.  And the "intended" will be changed to "shall" recently.  To avoid causing problems for other projects, we update the assembler to treat the architecture string "v_zvamo_zvlsseg" and "v" as the same for now.  That is, users only need to enable V extension, and then the vector AMO and vector load/store segment instructions are enabled, too.


gas/config/tc-riscv.c
```diff
@@ -142,17 +142,16 @@ riscv_multi_subset_supports (enum riscv_insn_class insn_class)
       return riscv_subset_supports ("f") && riscv_subset_supports ("c");
 
     case INSN_CLASS_Q: return riscv_subset_supports ("q");
-    case INSN_CLASS_V: return riscv_subset_supports ("v");
+
+    case INSN_CLASS_V:
+    case INSN_CLASS_V_AND_ZVLSSEG:
+      return riscv_subset_supports ("v");
```
jim-wilson, line 148: Maybe that should be
    case INSN_CLASS_V_AND_ZVLSSEG:
      return riscv_subset_supports ("v") || riscv_subset_supports ("zvlsseg");
which seems to be the intent of the V spec.  We probably don't have any way to enable zvlsseg without v but presumably this will be given a name later, and then when we add that this will work as intended.

And maybe the class name should be changed to INSN_CLASS_V_OR_ZVLSSEG?

Simiiarly with the zvamo support.

We can defer this to later if you think that is better though.

Nelson1225: How about simplify the Zv* class from INSN_CLASS_V_[AND|OR]_ZV* to INSN_CLASS_ZV*.  The naming INSN_CLASS_ZV* should clearly indicate that the instruction is a subset extension of V.  We only use the combination of naming when the instruction don't have a clear naming, like INSN_CLASS_F_AND_C and INSN_CLASS_V_AND_F, ...

One minor thing is that, are the zvamo and zvlsseg instructions possible to execute when we only enable the zvamo and zvlsseg subsets, but without enabling the base v?  Presumably we will have a Zvbase which means the instruction is belongs to base V, excepts zvamo and zvlsseg.  So we can change the code as follows,

case INSN_CLASS_V:
return riscv_subset_supports ("v")
or riscv_subset_supports ("zvbase");

case INSN_CLASS_ZVLSSEG:
return riscv_subset_supports ("v")
or (riscv_subset_supports ("zvbase") and riscv_subset_supports ("zvlsseg"));

case INSN_CLASS_ZVQMAC:
return (riscv_subset_supports ("v") and riscv_subset_supports ("zvqmac")
or (riscv_subset_supports ("zvbase") and riscv_subset_supports ("zvqmac"));

jim-wilson:
The way that I would expect this to work is if you have zvlsseg then you have the vector registers and csr and hardware, even if you don't have the base vector instructions.  So we don't need to worry about whether any v or zvbase has been selected.  zvbase means that you have the regular vector instructions like vadd and others.  So V is equivalent to zvbase and zvlsseg and zvamo.  If you have just vbase, then you don't specify v, you only specify zvbase.  Similarly, if you have just the vlsseg instructions, then you don't specify v you only specify zvlsseg.  Except that we don't have a name for zvbase yet so we can't implement that part.

See for instance how the B extension support works on the riscv-bitmanip branch.  I think the vector support should work the same way.  So the B extension has zbb which are the base bitmanip instructions, and you can enable them by either listing b or zbb in the arch option, which is handled by INSN_CLASS_B_OR_ZBB.  You can use B to enable all of the default parts of the B extension, or you can list the individual zb* extensions that you have, which does not necessarily have to include zbb.

Nelson1225:
> The way that I would expect this to work is if you have zvlsseg then you have the vector registers and csr and hardware, even if you don't have the base vector instructions.

I also update the riscv_csr_class_check, too.   We now can use the vector register by setting -mcsr-check and only enabling the zvlsseg and zvamo.

> So we don't need to worry about whether any v or zvbase has been selected. zvbase means that you have the regular vector instructions like vadd and others. So V is equivalent to zvbase and zvlsseg and zvamo. If you have just vbase, then you don't specify v, you only specify zvbase. Similarly, if you have just the vlsseg instructions, then you don't specify v you only specify zvlsseg. Except that we don't have a name for zvbase yet so we can't implement that part.

Thank you very much for clarifying this :)  Now I know how to handle these.

PR #220, opened 2020-07-03
Update rvv-1.0.x to the newest vector spec v1.0 released (Not yet have the release tag).

Nelson1225:
There are some changes according to the vector v1.0 released spec,

* Extend the VL<nf>R and VS<nf>R instruction, when <nf> is 2, 4, 8.
https://github.com/riscv/riscv-v-spec/commit/0fec0e257cca5e389a0e1774e7d5aaac729b4ad2#diff-4fa320b4e415eee7270590bb06607389R1923

* Added element width hint to whole register loads/stores. (Encode EEW for VL<nf>RE<EEW>)
https://github.com/riscv/riscv-v-spec/commit/20f673c9aef9ee2ee18a30db52b9a2c5d287deb5#diff-4fa320b4e415eee7270590bb06607389R1910

* Remove Vector indexed instruction when EEW >= 128.
https://github.com/riscv/riscv-v-spec/commit/0fec0e257cca5e389a0e1774e7d5aaac729b4ad2#diff-4fa320b4e415eee7270590bb06607389R1532

* Add assembly pseudo instructions vncvt.x.x.v.
https://github.com/riscv/riscv-v-spec/commit/2ab0b81ddc3476de734d1bfb065d04bb8730a36b#diff-4fa320b4e415eee7270590bb06607389R2682

* Added reciprocal and reciprocal square-root estimate instruction.
https://github.com/riscv/riscv-v-spec/commit/d35b23f1a2c2e8d2a13fba1fae52b84e75c6d483#diff-34bccafef6cfe01367796362310416dfR190

* Added vrgatherei16 instruction.
https://github.com/riscv/riscv-v-spec/commit/a67925038b653a75cd7eadb68b9915449941745d

* Make vlmul bits contiguous in vtype. (done in the https://github.com/riscv/riscv-binutils-gdb/pull/222)
https://github.com/riscv/riscv-v-spec/commit/b8cd98bc9467617b6653583116b2dfe9eefb2f51

* ...

Thanks
Nelson


include/opcode/riscv-opc.h
```diff
@@ -1557,8 +1557,21 @@ xxx - 00 x 10000 xxxxx --- xxxxx 0000111 VLSEG<nf>E<EEW>FF
 
 #define MATCH_VL1RV  0x02800007
 #define MASK_VL1RV   0xfff0707f
+#define MATCH_VL2RV  0x12800007
```
jim-wilson, line 1560: Is this right?  The <nf> field is 3 bits, so it looks like it should be 0x2....  And similarly below the 0x3 should be 0x6 and the 0x7 should be 0xe.

Nelson1225:
> Looks like quite a few recent changes. The vlXr now have an optional eX. There are new vfrece7 and vfrsqrte7 instructions. The index values are now specified to stop at ei64 where as we went up to ei1024. There is a new vncvt alias. Some of the source/dest overlap rules have changed. That is just what I noticed. There might also be other changes in there I missed.

OK I will start to add these today.

include/opcode/riscv-opc.h
Nelson1225: No it's wrong....  I’m sorry to have made such a stupid mistake....  Already fixed now.  And I also notice that we have a new mew encoding for vector load/store, these will be added by other patches.  Thanks.

Nelson1225:
The binutils 2.35 is released now, so I close this PR, and will create a new one to the rvv-1.0.x, which is based on the binutils 2.35.

PR #224, opened 2020-07-29
Update RVV v1.0 (until the vlmul change) based on the binutils-2.35

Nelson1225:
There are nine commits here,

1. RISC-V: Support RVV according to vector spec v1.0-draft-20200516.
This commit is ported from rvv-0.9.x branch.  Basically, this is the v0.9 implementation.

2. Extend the VL<nf>R and VS<nf>R instruction, when <nf> is 2, 4, 8. 
3. Remove Vector indexed instruction when EEW >= 128.
https://github.com/riscv/riscv-v-spec/commit/0fec0e257cca5e389a0e1774e7d5aaac729b4ad2#diff-4fa320b4e415eee7270590bb06607389R1923

4. Add assembly pseudo instructions vncvt.x.x.v.
https://github.com/riscv/riscv-v-spec/commit/2ab0b81ddc3476de734d1bfb065d04bb8730a36b

5. Added reciprocal and reciprocal square-root estimate instruction.
https://github.com/riscv/riscv-v-spec/commit/d35b23f1a2c2e8d2a13fba1fae52b84e75c6d483

6. Added element width hint to whole register loads/stores.
https://github.com/riscv/riscv-v-spec/commit/20f673c9aef9ee2ee18a30db52b9a2c5d287deb5

7. Added vrgatherei16 instruction.
https://github.com/riscv/riscv-v-spec/commit/a67925038b653a75cd7eadb68b9915449941745d

8. Make vlmul bits contiguous in vtype.
https://github.com/riscv/riscv-v-spec/commit/b8cd98bc9467617b6653583116b2dfe9eefb2f51

9. RISC-V: Clarify the supported versions for the unratified extensions.
Based on the recent changes in the binutils-2.35 (-misa-spec option), we probably need to set the draft versions for those UN-ratified extensions.  This commit will set the current v1.0 to the V and it's sub extensions correctly by default.  Note that only one default draft version can be accepted for an extension, since there is no ISA spec define them.  If the next draft is released, then we must remember to update the draft version in the riscv_ext_version_table.


jim-wilson:
This looks OK to me.

I think we missed a problem with dest/mask register overlaps when the mask length changed to 1 bit per element.  The v0.8 spec says "The destination vector register group for a masked vector instruction can only overlap the source mask register (v0) when LMUL=1. Otherwise, an illegal instruction exception is raised."  Since we can't know LMUL we don't check this in the assembler.  However, in v1.0 is says "NOTE: Some masked instructions that target `v0` which were legal in v0.8 are illegal with the new MLEN=1 mask layout for v1.0. For example, `vadd.vv v0, v1, v2, v0.m` is now always illegal; previously, it was legal for LMUL=1."  Since it is always illegal now, the assembler can generate an error.  It isn't clear to me why this is illegal though.  The text says mask reg has EEW of 1, and overlap with dest is OK if it is the highest reg of the set, for LMUL=1 v0 would be the highest reg of the set.  Maybe this rule only applies if dest LMUL is greater than 1?  In that case v0 can never be the highest reg of the group so this would always be illegal.

There seems to be a problem with how zvamo and zvlsseg are handled.  The current text says "This instruction subset is given the ISA string `Zvamo`.  NOTE: This set of instructions is included in the base "V" extension used for the Unix profile."  Note the reference to the unix profile.  This suggests that when the target is *-*-linux-gnu then 'v' should include zvamo, but when the target is *-*-elf it should not.  It isn't clear if this is what the V spec authors intended.  The zvlsseg language is the same.  Currently we make v imply zvamo unconditionally, and don't for zvlsseg.

I think these issues can be fixed with follow on patches, if necessary.

kito-cheng:
@aswaterman do you mind disambiguate for this?

> NOTE: This set of instructions is included in the base "V" extension used for the Unix profile
> This suggests that when the target is --linux-gnu then 'v' should include zvamo, but when the target is --elf it should not.
@aswaterman Does it what you mean? my understanding is V = zvbase + zvamo + zvlsseg no matter the target triple.

aswaterman:
@kito-cheng that's my understanding, too.  The problem is that no one ever formally defined the name Zvbase, so there's no way to refer to that piece; you should raise that issue on the vector spec.

Nelson1225:
> I think we missed a problem with dest/mask register overlaps when the mask length changed to 1 bit per element. The v0.8 spec says "The destination vector register group for a masked vector instruction can only overlap the source mask register (v0) when LMUL=1. Otherwise, an illegal instruction exception is raised." Since we can't know LMUL we don't check this in the assembler. However, in v1.0 is says "NOTE: Some masked instructions that target `v0` which were legal in v0.8 are illegal with the new MLEN=1 mask layout for v1.0. For example, `vadd.vv v0, v1, v2, v0.m` is now always illegal; previously, it was legal for LMUL=1." Since it is always illegal now, the assembler can generate an error. It isn't clear to me why this is illegal though. The text says mask reg has EEW of 1, and overlap with dest is OK if it is the highest reg of the set, for LMUL=1 v0 would be the highest reg of the set. Maybe this rule only applies if dest LMUL is greater than 1? In that case v0 can never be the highest reg of the group so this would always be illegal.

Thanks for the reminder, I will start to handle this next week.

Nelson1225:
I will start to update the mask overlap issue this week.  As for the Zvbase, we will create an issue on the vector spec to discuss it.  Thanks for all @jim-wilson @aswaterman @kito-cheng

kito-cheng:
Need sub-extension for baseline vector ISA: https://github.com/riscv/riscv-v-spec/issues/546
Clarify note for Zvamo and Zvlsseg: https://github.com/riscv/riscv-v-spec/issues/547

PR #227, opened 2020-08-14
Update mask and some missing constraints to RVV v1.0.

Nelson1225:
There are four commits as follows,

[PATCH-1]  RISC-V: Update the vector mask constraints.
For now, the destination vector register group for most of masked vector instruction can not overlap the source mask register.  Unless,

* Comparison instruction.
* Reduction instruction.

V0 is a special case.  It is a carry-in register for v[m]adc and v[m]sbc, and is used to choose vs1/rs1/frs1/imm or vs2 for v[f]merge. 
 For vadc and vsbc, vd can not overlap the carry-in V0.  I use the match_vd_neq_vm to check the constraints since V0 and VM have the same encoding. 

I regard vdot[u].vv as a reduction instruction, too.  Therefore, I don't check the VM constraints for it.

Besides, I also rewrite the vector failed test cases since their names are hard to understand and maintain.

[PATCH-2]  RISC-V: Add the missing constraints for VL<nf>R and VS<nf>R.
The destination vector register must be aligned to <nf>.  The <nf> is 1, 2, 4, 8 for now.  Besides, under the aligned constraints, it is impossible to use the registers that don't exist (number > 31).

[PATCH-3]  RISC-V: Fix the constraints for vector mask and compress instructions.
Add the missing constraints for vmsbf.m, vmsif.m and vmsof.m.  Their destination vector register can not overlap the source register.

Besides, we don't need to check the VM overlap for vcompress.vm, so change the match_func from match_vd_neq_vs1_neq_vs2 to match_vd_neq_vs1_neq_vs2_neq_vm.

[PATCH-4]  RISC-V: Report rvv assembler constraint errors in detail.
I think it would be helpful if assembler can report more detailed errors when parsing instructions.  I tried to improve them before, but still have some problems need to be resolved.  In general, we can check the constraints when parsing the operands, or in the `match_func` when the parsing is done.  However, reporting detailed rvv constraints messages in the `match_func` is relatively clear and simple without any potential problems.  Besides, most of the rvv constraints are checked in their `match_func`, so I only improve them for now, and it should be enough to handle most of cases.


Nelson1225:
@jim-wilson @kito-cheng Thanks for all!

PR #233, opened 2020-10-15
RISC-V: Support pause instruction as a hint fence w,0.

Nelson1225:
There is a problem here - Should we regard the pause instruction as an alias of hint fence?  Here is the assembly syntax of fence instruction,

fence  [i][o][r][w], [i][o][r][w]

The current assembler doesn't allow the pred and succ fields are set to `0`.  Therefore, if the pause is an alias of fence, then we will get `fence   w,unknown` when the option "-Mno-aliases" is set to objdump.

On the other hand, should we accept that the pred and succ fields can be set to `0` by the assembly fence instruction?  That is,

fence 0, w  --> illegal for now, should this be legal?
fence w, 0  --> same as above
fence 0, 0  --> same as above

If it is accepted, then it's fine to regard the pause instruction as an alias.

However, this patch doesn't regard the pause as an alias, and the `0` pred and succ are not accepted for the assembly fence.

### DIFF
--- a/gas/testsuite/gas/riscv/fence-tso.s
+++ b/gas/testsuite/gas/riscv/fence-tso.s
@@ -1,2 +1,0 @@
-target:
-	fence.tso
--- a/gas/testsuite/gas/riscv/fence.d
+++ b/gas/testsuite/gas/riscv/fence.d
@@ -8,4 +8,6 @@ Disassembly of section .text:
 
 0+000 <target>:
 [ 	]+0:[ 	]+8330000f[ 	]+fence.tso
-
+[ 	]+4:[ 	]+0100000f[ 	]+pause
+[ 	]+8:[ 	]+0ff0000f[ 	]+fence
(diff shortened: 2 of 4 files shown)


Nelson1225:
Here is the ISA spec PR, https://github.com/riscv/riscv-isa-manual/pull/398

aswaterman:
IMO, we should always have allowed pred=0/succ=0; that was an oversight on my part.

I probably would have used `-` as the syntax instead of `0`, but either works.

kito-cheng:
> IMO, we should always have allowed pred=0/succ=0; that was an oversight on my part.
> 
> I probably would have used `-` as the syntax instead of `0`, but either works.

Both binutils and LLVM are only support `iorw` as operands only, maybe we should also update https://github.com/riscv/riscv-asm-manual too?

Nelson1225:
We don't need this anymore, so I delete the PR.

PR #236, opened 2020-11-05
Allow more verbosity when reaching invalid opcode while prologue scan

AntonKrug:
Handle the invalid instructions slightly differently from the ignored/other instructions

- Stop scanning in both cases
- In case of invalid instruction do verbose printf when the 'riscv_debug_unwinder' is set true
- While the ignored/other instructions will stop the scanner quietly no matter what 'riscv_debug_unwinder' is set to

### DIFF
--- a/gdb/riscv-tdep.c
+++ b/gdb/riscv-tdep.c
@@ -1126,6 +1126,10 @@ public:
       LR,
       SC,
 
+      /* instruction opcodes with zeros/ones indicating  that the scan might
+	 be executed on a unimplemented memory */
+      INVALID,
+
       /* Other instructions are not interesting during the prologue scan, and
 	 are ignored.  */
       OTHER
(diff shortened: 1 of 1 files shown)


jim-wilson:
Patches must be sent upstream, and GNU toolchain patches require FSF copyright assignments.  Unless perhaps they are trivial and not copyrightable, but there are limits on how many such patches can be accepted without copyright assignments.  This would be an issue for the gdb maintainers to discuss.  You can send gdb patches to gdb-patches@sourceware.org but if you don't already have a copyright assignment, it might be better to just submit a bug report, and let someone who has an assignment write the necessary patch.  If you are interesting in making further contributions to the GNU toolchain, then you should look into getting a copyright assignment.  You can start the process by sending email to assign@gnu.org and mentioning what you want to contribute to.  A full toolchain assignment would include binutils, gdb, gcc, and glibc for instance.  If you have an employer who can claim copyright to the code you write, then you need a signoff from a VP or higher, which depending on who you work for, may or may not take some time, and may or may not be possible.

AntonKrug:
Thank you very much Jim, your comment (as always) is very helpful. Will try to get them under FSF as they are fairly trivial, but getting assignment would be great for the future.

PR #238, opened 2020-11-23
Update two features for rvv-1.0.x.

Nelson1225:
1. Add assembly pseudoinstructions, vneg.v and vfneg.v.
2. Add ordered/unordered vector indexed instructions.


jim-wilson (approved):

jim-wilson:
Reviewing the V spec, I noticed that the zvqmacc and zvediv subextensions are not part of the v1.0 spec, and their instructions were removed from the opcode table, so maybe we shouldn't be implementing these by default?  We could just put #if 0 around instructions in the riscv-opc.c file for instance.  Or we could remove everything, the MASK/MATCH macros, the zvediv zvqmacc extensions, and the instructions in the riscv-opc.c file.  I also noticed that the v*dot* instructions are not marked as part of the zvediv subextension.  That definintely needs to be fixed.  Otherwise we will be supporting them as part of the base V extension. 
 This can all be fixed in a follow on patch.  This one is big enough already.

Nelson1225:
> Reviewing the V spec, I noticed that the zvqmacc and zvediv subextensions are not part of the v1.0 spec, and their instructions were removed from the opcode table, so maybe we shouldn't be implementing these by default? We could just put #if 0 around instructions in the riscv-opc.c file for instance. Or we could remove everything, the MASK/MATCH macros, the zvediv zvqmacc extensions, and the instructions in the riscv-opc.c file.

#if 0 is good to me.  Otherwise, we probably need to find a place to put the removed zvqmacc and zvediv extensions.  I think we should avoid to create a new branch in riscv/sifive...  Therefore, put them in my GitHub repo should be fine, if we decide to remove the code directly.  However, thanks for the #if 0 suggestion, personally I prefer to use it :)

> I also noticed that the v_dot_ instructions are not marked as part of the zvediv subextension. That definintely needs to be fixed.  Otherwise we will be supporting them as part of the base V extension.  This can all be fixed in a follow on patch. This one is big enough already.

Got it, I will fix them in the future patches.

PR #241, opened 2020-12-12
Update rvb binutils support.

jim-wilson:
Add zext.[bhw] that gcc now needs.  Remove dropped insns.  Fix insns that have
changed subsets.  Fix max/minu opcodes.  Add new xperm.[nbhw] insns.  Minimally
verfified.


Nelson1225:
LGTM 👍

PR #247, opened 2021-02-24
Back to v1.0-draft-20210130, and then added the pseudo vfabs.v.

Nelson1225:
Updated the spec version to the v1.0-draft-20210130, and then added the pseudo vfabs.v as the alias of vfsgnjx.vv.


jim-wilson (approved):

Nelson1225:
Merged, thanks.

PR #249, opened 2021-02-25
Changes required to build gdb.

jim-wilson:
Use _ instead of . in DECLARE_INSN since we can't put . in the middle of a function name.


Nelson1225:
LGTM, thanks.

PR #252, opened 2021-03-10
Fix match value for 'xperm.h'

### LINKED ISSUE #251: riscv-binutils-2.35-rvb: wrong match for xperm.h
Hello,

at: https://github.com/riscv/riscv-binutils-gdb/blob/3e5f50f31348d05144966545d862a3172d315230/include/opcode/riscv-opc.h#L378

The match value is wrong and should be:
`#define MATCH_XPERMH 0x28006033` (6 instead of 3 for bits 15-12).

as according to the current bitmanip draft, funct3 for xperm.h is `110` (and not `011`). The other forms (`.b`, `.n` and `.w`) are correct.

rdolbeau:
closes #251


Nelson1225:
LGTM.

PR #253, opened 2021-03-23
Backport fix from FSF to fix configure problem caused by ctf patches.

jim-wilson:
Including bfd.h now also requires including string.h.  Fixes riscv-gnu-toolchain #867.


jim-wilson:
Verified by building with CFLAGS="-O2 -g -Werror=implicit-function-declaration" with and without the patch.  Without the patch gdb fails to read ELF files, with the patch gdb can read ELF files.

kito-cheng:
Thanks :)

PR #254, opened 2021-03-30
Add k-ext support with v1.0.0-rc

pz9115:
PLCT had add k-ext support with crypto spec v0.90 with finish all regress testing in binutils part, please check it.


Nelson1225:
Shouldn't these be sent to riscv-binutils-experiment branch?

jim-wilson (commented):
Overall it looks OK.  There are a just a few minor questions I have, and some apparent issues with changing the patch base which confused the patch a little.

jim-wilson:
In general, I would prefer that patches go upstream.  Branches here in github.com/riscv/riscv-binutils-gdb are becoming a serious liability.  But we are still in the process of moving stuff upstream (waiting for gcc-11 branch), and using github for patch reviews is probably more convenient than email for many people.
]

jim-wilson:
Kito's review of the K gcc patch reminds me, I didn't see testcases.  It is generally a good idea to add a gas testcase for new extensions, to verify that we can assemble all of the instructions, and then disassemble them correctly.

We will also need to make sure we didn't accidentally break gdb.  We can't do that in riscv-gnu-toolchain because we use different release branches for binutils and gdb.  But in the FSF development tree, binutils and gdb development sources are on the same master branch.

bfd/elfxx-riscv.c
```diff
@@ -1159,8 +1159,17 @@ static bfd_boolean
...
-      || strcmp (subset, "zifencei") == 0)
+      || strcmp (subset, "zifencei") == 0
+      || strcmp (subset, "zkg") == 0
+      || strcmp (subset, "zkb") == 0
+      || strcmp (subset, "zkr") == 0
+      || strcmp (subset, "zkne") == 0
+      || strcmp (subset, "zknd") == 0
+      || strcmp (subset, "zknh") == 0
+      || strcmp (subset, "zksed") == 0
+      || strcmp (subset, "zksh") == 0)
```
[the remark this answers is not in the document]
Nelson1225: Yeah k and it's sub-extensions should have a draft version.  We won't output the extensions in the output ISA string, if you list them in the riscv_ext_dont_care_version. The function used to deal with the special extensions like zicsr and zifencei. Since they are extracted from i in the ISA spec v2.1, we should add them as the implicit extensions when the version of i is less than v2.1, but in the newer spec they are explicit. For example, add zicsr as implicit means we allow to use the related csrr instructions when only enable i v2.0, but we won't output them to the output ISA string since they don't have a default versions. The g-ext is another special case that it's version doesn't mean anything.

bfd/elfxx-riscv.c
```diff
@@ -1315,7 +1324,7 @@ riscv_parsing_subset_version (riscv_parse_subset_t *rps,
 const char *
 riscv_supported_std_ext (void)
 {
-  return "mafdqlcbjtpvn";
+  return "mafdqlcbjktpvn";
```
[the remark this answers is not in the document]
Nelson1225: I can not find the order of k either.  The orders of extensions used to be listed in the Figure 27.1 in the ISA spec.  It seems like the draft spec doesn't mention k and it's order.

bfd/elfxx-riscv.c
```diff
@@ -1760,6 +1769,69 @@ riscv_parse_add_implicit_subsets (riscv_parse_subset_t *rps)
...
+      riscv_parse_add_subset (rps, "zknh",
+                              RISCV_UNKNOWN_VERSION,
+                              RISCV_UNKNOWN_VERSION, TRUE);
+      riscv_parse_add_subset (rps, "zksed",
+                              RISCV_UNKNOWN_VERSION,
+                              RISCV_UNKNOWN_VERSION, TRUE);
+      riscv_parse_add_subset (rps, "zksh",
+                              RISCV_UNKNOWN_VERSION,
+                              RISCV_UNKNOWN_VERSION, TRUE);
+      riscv_parse_add_subset (rps, "zknh",
```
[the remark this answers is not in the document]
Nelson1225: I'm a little confused about the imply extensions.  For g, it will be expanded to imafd, and won't be generated to the output ISA string.  For v, we won't add the implicit zvamo and zvlsseg, we use INSN_CLASS_V_OR_VAMO for instead.  It means we won't add the zvamo and zvlsseg to the ISA output ISA string when only enable v, but we can use the instructions.  Maybe it’s because v didn’t have the implicit implementation at that time.  It might be worth taking some time to clarify, I think the better rules are as follows,

* If the base extension can be divided into separate Z extensions, and be covered by them, then I prefer to think of the base extension as g.  We will add multiple implicit Z extensions for it, and then only output the Z extensions to the output ISA string.  I think b is just like this example, and k maybe also is.  I haven't see the spec of k in details, so not sure about this.

* Some of v instructions don't belong to the Zv* extensions, they belongs to base v directly, so the base v cannot be suppressed.  For those zvamo instructions, like vamoaddei8.v, we probably can change their instruction classes from INSN_CLASS_V_OR_ZVAMO to INSN_CLASS_ZVAMO, and add them as v's implicit extensions.  Therefore, we can only enable base v, and then can use zvamo and zvlsseg instructions.  Also output them to the output ISA string with the default versions.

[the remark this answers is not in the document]
jim-wilson: Yes, K is like B.  Everything is in a zk* subset, and K is just a collection of zk* subsets.  We probably should canonicalize like we do for g/imafd.

V is confused.  The zvamo chapter says "This set of instructions is included in the base "V" extension used for the Unix profile." which implies that zvamo is part of V.  But if zvamo is part of V, then there would be no reason to have a name for it.  And that leads to the question of whether zvamo part of V for embedded targets.  We have an open issue for this, riscv-v-spec 548.  Krste's proposed solution in riscv-v-spec issue 550 is to add a whole bunch of zve* extensions for embedded, in which case the zvamo name becomes useful again.  I would suggest not trying to make any generalizations based on how V works, as V is still too confused over its subextension naming scheme.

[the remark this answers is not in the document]
Nelson1225: > Yes, K is like B. Everything is in a zk* subset, and K is just a collection of zk* subsets. We probably should canonicalize like we do for g/imafd.

Thanks, this looks reasonable to me :)

> V is confused. The zvamo chapter says "This set of instructions is included in the base "V" extension used for the Unix profile." which implies that zvamo is part of V. But if zvamo is part of V, then there would be no reason to have a name for it. And that leads to the question of whether zvamo part of V for embedded targets. We have an open issue for this, riscv-v-spec 548. Krste's proposed solution in riscv-v-spec issue 550 is to add a whole bunch of zve* extensions for embedded, in which case the zvamo name becomes useful again. I would suggest not trying to make any generalizations based on how V works, as V is still too confused over its subextension naming scheme.

OK, so I think keep the current implementations until the vector spec clarify them is a wise choice.

include/opcode/riscv-opc.h
```diff
@@ -621,7 +621,121 @@
...
+#define MATCH_CUSTOM3_RS1 0x207b
+#define MASK_CUSTOM3_RS1  0x707f
+#define MATCH_CUSTOM3_RS1_RS2 0x307b
+#define MASK_CUSTOM3_RS1_RS2  0x707f
+#define MATCH_CUSTOM3_RD 0x407b
+#define MASK_CUSTOM3_RD  0x707f
+#define MATCH_CUSTOM3_RD_RS1 0x607b
+#define MASK_CUSTOM3_RD_RS1  0x707f
+#define MATCH_CUSTOM3_RD_RS1_RS2 0x707b
+#define MASK_CUSTOM3_RD_RS1_RS2  0x707f
```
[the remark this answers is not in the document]
pz9115: Right, I will delete all of them

bfd/elfxx-riscv.c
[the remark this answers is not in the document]
jim-wilson: I opened riscv-isa-manual issue 636 asking for this to be fixed, and suggested this solution.

bfd/elfxx-riscv.c
```diff
@@ -1027,6 +1027,8 @@ riscv_elf_add_sub_reloc (bfd *abfd,
 }
 
 #define RISCV_UNKNOWN_VERSION -1
+#define RISCV_CRYPTO_MAIN_VERSION 9
+#define RISCV_CRYPTO_SUB_VERSION 0
```
Nelson1225, line 1031: Please define the default version in the ext_version_table rather than here,
https://github.com/riscv/riscv-binutils-gdb/blob/riscv-binutils-experiment/gas/config/tc-riscv.c#L100

bfd/elfxx-riscv.c
```diff
@@ -1760,6 +1761,66 @@ riscv_parse_add_implicit_subsets (riscv_parse_subset_t *rps)
     riscv_parse_add_subset (rps, "zicsr",
 			    RISCV_UNKNOWN_VERSION,
 			    RISCV_UNKNOWN_VERSION, TRUE);
+  else if ((riscv_lookup_subset (rps->subset_list, "k", &subset)))
```
Nelson1225, line 1764: I suppose k, zkn and zks are similar to g, they used to expand into others.  We won't output g to the final architecture string, so k, zkn and zks should be the same.  I had done some improvement on FSF binutils about this, so it might be easier to complete over there.

Nelson1225: However,my old implementation here did not consider so much, so it looks OK.

gas/config/tc-riscv.c
```diff
@@ -143,6 +143,18 @@ static const struct riscv_ext_version ext_version_table[] =
   {"zba",   ISA_SPEC_CLASS_DRAFT, 0, 93},
   {"zbc",   ISA_SPEC_CLASS_DRAFT, 0, 93},
 
+  {"k",   ISA_SPEC_CLASS_DRAFT, 0, 90},
```
Nelson1225, line 146: Same as above, if k, zkn and zks are expansions, then we don't need to add the default versions for them.

gas/config/tc-riscv.c
```diff
@@ -2693,6 +2724,28 @@ riscv_ip (char *str, struct riscv_cl_insn *ip, expressionS *imm_expr,
 			    "specifier `F%c'\n"), *args);
 		}
 	      break;
+            
+	    case 'y':        /* bs immediate */
+	            my_getExpression (imm_expr, s);
```
Nelson1225, line 2729: Wrong GNU indent.

bfd/elfxx-riscv.c
```diff
@@ -1760,6 +1761,66 @@ riscv_parse_add_implicit_subsets (riscv_parse_subset_t *rps)
...
+    {
+      riscv_parse_add_subset (rps, "zkg",
+                              RISCV_UNKNOWN_VERSION,
```
Nelson1225, line 1767: Wrong GNU indent.

include/opcode/riscv-opc.h
```diff
@@ -867,6 +934,7 @@
 #define CSR_TINFO 0x7a4
 #define CSR_TCONTROL 0x7a5
 #define CSR_MCONTEXT 0x7a8
+#define CSR_MNOISE 0x7a9
```
Nelson1225, line 937: Seems like you add new CRSs, but forgot to declare them by DECLARE_CSR.

Nelson1225: Besides, add new CSR should update the series of priv-reg* test cases.

include/opcode/riscv-opc.h
```diff
@@ -621,7 +621,73 @@
 #define MASK_C_LDSP  0xe003
 #define MATCH_C_SDSP 0xe002
 #define MASK_C_SDSP  0xe003
+
+/* K-ext*/
```
Nelson1225, line 625: GNU comment style.

Nelson1225 (changes requested):
Generally looks good, except the k expansions, GNU coding standards and CSR issue.  We don't need to care the k expansions here, we can fix them when sending the draft k stuff to the FSF users/riscv/binutils-integration-branch.

However, we should all pay attention to the GNU coding standards.  You can refer to my previous three patches as follows,

* RISC-V: Comments tidy and improvement.
* RISC-V: Error and warning messages tidy.
* RISC-V: Indent and GNU coding standards tidy, also aligned the code.

After fixing the GNU comments and indents, and the CSR problem, the patch looks good to me.

Nelson1225:
FYI, you can press the "Resolve conversation" if you have fixed them.  So that the PR will look a little cleaner, thanks.

bfd/elfxx-riscv.c
pz9115: Okay, if it need change, I will update it.

gas/config/tc-riscv.c
pz9115: Okay, I deleted the version for k,zkn,zks.

gas/config/tc-riscv.c
pz9115: Sorry for the indent problems, fixed.

bfd/elfxx-riscv.c
pz9115: Sorry for the indent problems, fixed.

include/opcode/riscv-opc.h
pz9115: Thanks for the point out, I had added it.

include/opcode/riscv-opc.h
pz9115: Fixed it. Thanks.

gas/config/tc-riscv.c
Nelson1225: Sorry that may have misunderstood you.  I meant this is OK for riscv-binutils-experiment, but could be improved in the FSF binutils.  Therefore, we don't need to care about the k and p's expansions in the riscv-binutils-experiment.

include/opcode/riscv-opc.h
```diff
@@ -710,6 +776,7 @@
 #define CSR_MARCHID 0xf12
 #define CSR_MIMPID 0xf13
 #define CSR_MHARTID 0xf14
+#define CSR_MENTROPY 0xf15
```
Nelson1225, line 779: Maybe you forgot this one :)

Nelson1225 (approved):
LGTM, thanks.

gas/config/tc-riscv.c
pz9115: Ok, let me delete the expansions with k,zkn,zks in the riscv-binutils-experiment..

include/opcode/riscv-opc.h
pz9115: Thank you for your metion, I had update the 'mentorpy' case.

pz9115:
We had updated with the scalar crypto spec 1.0.0-rc, and merge it with the b-ext spec 1.0.0-rc contants zba_zbb_zbc_zbs for reusing  of zbk[bcx]. Sorry for the miss closing of this PR due to branch rename.

pz9115:
This work had been merged into the binutils upstream branch

https://sourceware.org/git/?p=binutils-gdb.git

close the PR.

PR #259, opened 2021-05-03
Fix crash when disassembling non-ELF files

JobNoorman:
The return value of get_elf_backend_data was used without checking for
NULL. Since this function seems to return NULL for non-ELF files, it
caused a crash when disassembling binary files. This commit adds a
NULL-check to fix this.


jim-wilson:
Patches must go upstream, and you must have an FSF copyright assignment in order for us to accept patches unless the patch is trivial.  Or you can submit a bug report, and let someone with an assignment write the patch.  See sourceware.org/binutils.

Nelson1225:
Already include this on upstream.

